PR #4, opened 2020-07-08
Integrate Debian patches

### LINKED ISSUE #2: Install locations and features missed on Debian
In pypa/setuptools#2232, we learned that Ubuntu has a customized version of distutils, so when Setuptools attempts to adopt distutils from the standard library, those customizations are lost and Ubuntu users are unhappy.

I propose to fold ~~[those patches (3.8)](https://sources.debian.org/patches/python3.8/3.8.4%7Erc1-1/distutils-install-layout.diff/)~~ [those patches](https://sources.debian.org/patches/python3.9/3.9.2-1/distutils-install-layout.diff/) into this project, enabling Setuptools to incorporate that functionality. This approach will need to be done in a way that only takes effect selectively on those systems.

jaraco:
Closes #2.

Tests pass, but what could go wrong?


jaraco:
I've done my best in f945849 and f65fe9d to restore the status-quo expectation on Debian and non-Debian systems. Assuming this approach is correct, it addresses pypa/setuptools#2232, unifies the implementation across all (?) platforms, and opens up the possibility of later adopting the Debian approach on non-Debian Linux systems.

jaraco:
Although previously merged, I've backed out this PR by force-pushing master to bbe8e80. Ref pypa/setuptools#2240.

PR #21, opened 2020-11-24
Do not use global cache in mkpath

dolfinus:
There is a global cache of created folders used by `distutils.dir_util.mkpath` and `distutils.dir_util.remove_tree` functions.

It is OK if only one command is run, and it uses first function for creating directory and second for removing it.

But it is definitely not OK in such cases as:
* Folder was created using `mkpath` but removed using `os.remove` or `shutil.rmtree`
* Folder was removed by some external process

In this PR I've removed this global cache to avoid such issues.

### DIFF
--- a/distutils/dir_util.py
+++ b/distutils/dir_util.py
@@ -7,10 +7,6 @@ import errno
 from distutils.errors import DistutilsFileError, DistutilsInternalError
 from distutils import log
 
-# cache for by mkpath() -- in addition to cheapening redundant calls,
-# eliminates redundant "creating /foo/bar/baz" messages in dry-run mode
-_path_created = {}
-
 # I don't use os.makedirs because a) it's new to Python 1.5.2, and
 # b) it blows up if the directory already exists (I want to silently
 # succeed in that case).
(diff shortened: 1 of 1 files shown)


dolfinus:
Hi.

Any updates here?

jaraco:
Thanks for the contrib. Unfortunately, distutils is a deeply rooted project depended on by a vast array of environments and configurations. This change largely removes behavior, behavior that is likely relied upon by some of those environments.

I appreciate that the change seems obvious to you, but it's not obvious to me this change is the best course of action or what implications it will have. Moreover, the description doesn't even include any description of what conditions where the issue was encountered. Is the issue theoretical or practical? Why not simply avoid the implicated uses described above?

More importantly, distutils is still in a state of flux. It hasn't been adopted by default for Setuptools yet, and until that happens, this project aims to restrict itself to only essential features and critical bugfixes.

Still, I wish not to lose track of this concern. If you still have interest in driving this change, please file an issue with the project and attempt to capture the larger story (what is the issue, who does it affect, how could it be fixed, what users will that impact, etc.).

Thanks again and good luck.

PR #44, opened 2021-06-23
Avoid symlink infinite loops

ssbarnea:
Related: https://github.com/pypa/setuptools/issues/332
Related: https://bugs.python.org/issue44497

### DIFF
--- a/distutils/filelist.py
+++ b/distutils/filelist.py
@@ -247,12 +247,22 @@ def _find_all_simple(path):
     """
     Find all files under 'path'
     """
-    results = (
-        os.path.join(base, file)
-        for base, dirs, files in os.walk(path, followlinks=True)
-        for file in files
-    )
-    return filter(os.path.isfile, results)
+    # https://bugs.python.org/issue44497
+    dirs = set()
+    files = set()
+    for dirpath, dirnames, filenames in os.walk(path, followlinks=True):
+        st = os.stat(dirpath)
+        scandirs = []
+        for dirname in dirnames:
+            st = os.stat(os.path.join(dirpath, dirname))
+            dirkey = st.st_dev, st.st_ino
+            if dirkey not in dirs:
+                dirs.add(dirkey)
+                scandirs.append(dirname)
+        dirnames[:] = scandirs
+        for f in filenames:
+            files.add(os.path.join(dirpath, f))
+    return files
 
 
 def findall(dir=os.curdir):


FFY00 (commented):
This will probably have a easier time getting in with tests 😉

distutils/filelist.py
```diff
@@ -247,12 +247,22 @@ def _find_all_simple(path):
...
+        scandirs = []
+        for dirname in dirnames:
+            st = os.stat(os.path.join(dirpath, dirname))
+            dirkey = st.st_dev, st.st_ino
+            if dirkey not in dirs:
+                dirs.add(dirkey)
+                scandirs.append(dirname)
+        dirnames[:] = scandirs
+        for f in filenames:
+            files.add(os.path.join(dirpath, f))
```
jaraco, line 264: Although it's a common pattern, I'm very much a detractor to the `init/loop/append` pattern (mostly because of how it mutates a local variable, making state analysis much more complicated).

The proposed approach entwines all of the logic of path traversal, link resolution, and deduplication into several loops.

Additionally, this change removes the check for `os.path.isfile`. I'm not sure that check is important, but it seems to be important enough that it was included previously.

I'd much rather see a solution that separates the concern of walking a tree while avoiding loops and assembling a list of files in that walk. Since distutils is only needed for Python 3.6+, perhaps a pathlib-based solution could help.

jaraco:
I'm unsure if the path walker should suppress duplicates due to loops or if it should fail in that situation. Why is it better to ignore the loops rather than report an error that there exists a circular loop?

I wonder too why distutils is special here. How is it that `walk` isn't solving this issue for other use-cases?

distutils/filelist.py
jaraco: I've got an alternative solution I'll present.

ssbarnea:
Circular links are perfectly fine and there are cases where they cannot be avoided. Their existence is not an "error", the code should be resilient and avoid going into endless loops. This can be avoided by normalising (resolving) the paths.

jaraco:
In 88f5c729bd, I took a stab at creating a test to repro the error and to serve as a regression test for any fix. But interestingly, I couldn't get os.walk to go into infinite recursion. As I dug into it, I found that `next(os.scandir(long))` after about 33 levels of indirection would raise an OSError:

```
(Pdb) entry, = os.scandir(long)       
(Pdb) entry
<DirEntry 'link-to-parent'>
(Pdb) entry.is_dir()
*** OSError: [Errno 62] Too many levels of symbolic links: '/private/var/folders/c6/v7hnmq453xb6p2dbz1gqc6rr0000gn/T/tmp1yzcz5nt/link-to-parent/link-to-parent/link-to-parent/link-to-parent/link-to-parent/link-to-parent/link-to-parent/link-to-parent/link-to-parent/link-to-parent/link-to-parent/link-to-parent/link-to-parent/link-to-parent/link-to-parent/link-to-parent/link-to-parent/link-to-parent/link-to-parent/link-to-parent/link-to-parent/link-to-parent/link-to-parent/link-to-parent/link-to-parent/link-to-parent/link-to-parent/link-to-parent/link-to-parent/link-to-parent/link-to-parent/link-to-parent/link-to-parent'
```

So at least on macOS, the reported infinite recursion never occurs, and as long as the symlink isn't pointing to some location that's expensive to traverse (like '/'), everything behaves nicely. Similarly on Ubuntu, I ran the same test and it completes quickly.

So it's going to prove difficult to have a reliable repro for this issue, as the reported issue is more complex than simple infinite recursion.

jaraco:
In 43fa2144bfcfb47c15d1d323d7c855c883bc510e, I found a way to assert that recursion and duplication isn't happening. This PR is obviated by #46.

PR #51, opened 2021-09-05
adopt more standard schema for pypy3.8

mattip:
PyPy 3.8 will use the standard posix schema as much as possible. In previous versions, downstream packagers like conda and linux distros had to patch this out anyway, this should make their lives easier. We cannot change the PyPy versions already released, so this is 3.8+ only. I intend to release PyPy 3.8 in a few weeks, so if this is correct it would be nice to get it out by then. 

In general, I think it would be beneficial to use the stdlib `sysconfig` where possible instead of repeating the code in `distutils.sysconfig`, it seems #23 is a step in this direction.

@mgorny, @isuruf

### DIFF
--- a/distutils/sysconfig.py
+++ b/distutils/sysconfig.py
@@ -100,11 +100,8 @@ def get_python_inc(plat_specific=0, prefix=None):
     if prefix is None:
         prefix = plat_specific and BASE_EXEC_PREFIX or BASE_PREFIX
     if os.name == "posix":
-        implementation = 'python'
-        if IS_PYPY:
-            if sys.version_info < (3, 8):
-                return os.path.join(prefix, 'include')
-            implementation = 'pypy'
+        if IS_PYPY and sys.version_info < (3, 8):
+            return os.path.join(prefix, 'include')
         if python_build:
             # Assume the executable is in the build directory.  The
             # pyconfig.h file should be in the same directory.  Since
@@ -116,6 +113,7 @@ def get_python_inc(plat_specific=0, prefix=None):
             else:
                 incdir = os.path.join(get_config_var('srcdir'), 'Include')
                 return os.path.normpath(incdir)
+        implementation = 'pypy' if IS_PYPY else 'python'
         python_dir = implementation + get_python_version() + build_flags
         return os.path.join(prefix, "include", python_dir)
     elif os.name == "nt":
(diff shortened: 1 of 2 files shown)


mattip:
CI is passing. It would be nice to move this forward since this is required once PyPy releases a python3.8 version

distutils/sysconfig.py
```diff
@@ -99,9 +99,12 @@ def get_python_inc(plat_specific=0, prefix=None):
...
         prefix = plat_specific and BASE_EXEC_PREFIX or BASE_PREFIX
-    if IS_PYPY:
-        return os.path.join(prefix, 'include')
-    elif os.name == "posix":
+    if os.name == "posix":
+        implementation = 'python'
+        if IS_PYPY:
+            if sys.version_info < (3, 8):
+                return os.path.join(prefix, 'include')
+            implementation = 'pypy'
```
FFY00, line 107:
```suggestion
        implementation = 'pypy' if IS_PYPY else 'python'
        if IS_PYPY and sys.version_info < (3, 8):
            return os.path.join(prefix, 'include')
```

distutils/sysconfig.py
```diff
@@ -142,13 +145,17 @@ def get_python_lib(plat_specific=0, standard_lib=0, prefix=None):
...
-            return os.path.join(prefix, "lib-python", sys.version[0])
-        return os.path.join(prefix, 'site-packages')
+        if sys.version_info < (3, 8):
+            # PyPy-specific schema
+            if prefix is None:
+                prefix = PREFIX
+            if standard_lib:
+                return os.path.join(prefix, "lib-python", sys.version[0])
+            return os.path.join(prefix, 'site-packages')
+        implementation = 'pypy'
```
FFY00, line 158: Same here:

```python
implementation = 'pypy' if IS_PYPY else 'python'
if IS_PYPY and sys.version_info < (3, 8):
    # PyPy-specific schema
    if prefix is None:
        prefix = PREFIX
    if standard_lib:
        return os.path.join(prefix, "lib-python", sys.version[0])
    return os.path.join(prefix, 'site-packages')
```

(github does not let me make code suggestions in this location)

FFY00 (approved):
Just some minor nits. They couple the logic together and have a slight line count reduction.

Otherwise, everything looks great.

distutils/sysconfig.py
mattip: adopted

distutils/sysconfig.py
mattip: adopted

mattip:
Thanks for the review

PR #52, opened 2021-09-05
Prefer sysconfig.python_build

longnguyen2004:
Change the condition to `not sysconfig.python_build` to be consistent with
https://github.com/pypa/distutils/blob/aede9e035dde34b0f2a6e9fe3a3bd87d4b9d7370/distutils/command/build_ext.py#L235-L241

Cygwin maintainers: Is this behaviorally correct?

### DIFF
--- a/distutils/command/build_ext.py
+++ b/distutils/command/build_ext.py
@@ -220,7 +220,7 @@ class build_ext(Command):
         # For extensions under Cygwin, Python's library directory must be
         # appended to library_dirs
         if sys.platform[:6] == 'cygwin':
-            if sys.executable.startswith(os.path.join(sys.exec_prefix, "bin")):
+            if not sysconfig.python_build:
                 # building third party extensions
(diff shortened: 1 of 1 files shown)


FFY00 (approved):

FFY00:
I would still like to get a look from a cygwin maintainer if possible, but this looks correct to me. @jaraco who can we ping?

jaraco:
I don't have a specific contact for Cygwin. I welcome FFY00 or others to engage with that community and establish a relationship with a maintainer there, but otherwise, my inclination is to do the best we can from the contributors who are active.

jaraco (approved):
This change looks good and acceptable.

Since this change was implied by a change to another section of code, I'd like to explore if those common elements can be refactored to be unified. I'll explore that possibility, but otherwise have no concerns with the approach here.

jaraco:
After reviewing the code, I'm not confident a refactor can be had that's safe without losing readability, so I'm going to accept this PR as proposed.

PR #59, opened 2021-09-26
bdist_rpm: add option to force version

jaraco:
The RPM versioning guidelines are slightly at odds with PEP440.
For example, RPM considers 1.1a1 to be a higher version than 1.1
where as PEP440 shows 1.1a1 to be an alpha and 1.1 to be the
official release.
Permitting specific override of the RPM version will allow a
maintainer to explicitly correct the version to fit their needs.

Originally submitted as pypa/setuptools#2778

### DIFF
--- a/distutils/command/bdist_rpm.py
+++ b/distutils/command/bdist_rpm.py
@@ -120,6 +120,10 @@ class bdist_rpm(Command):
         ('force-arch=', None,
          "Force an architecture onto the RPM build process"),
 
+        # Allow a packager to explicitly force a version
+        ('force-version=', None,
+         "Force a version onto the RPM instead of using metadata"),
+
         ('quiet', 'q',
          "Run the INSTALL phase of RPM building in quiet mode"),
         ]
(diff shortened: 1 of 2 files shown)


jaraco:
@jcpunk Can you investigate why the tests are failing?

jcpunk:
This is very puzzling.... `force_version` is getting set in the test `setup.py`.  However, it doesn't seem like anything I set under the `options: { 'bdist_rpm':` section gets passed into the unit tests.

I tested a number of existing options and those also are ignored.  The most obvious are `release` or `build_requires` which don't propagate down...

I'll confess I'm not sure how to dig deeper into that....

jaraco:
I suggest to dive into the code and get more familiar with the architecture of distutils. I think the problem with the approach is you're creating a new option to the `bdist_rpm` command, but in the tests, you're supplying new keyword arguments to the distribution. Take a look at `test_quiet` to see how that test exercises the `--quiet` option to `bdist_rpm`.

jaraco:
Feel free to grab this branch into your own repo and iterate on it there. Since you'll probably need to create a new PR, I'll close this one.

PR #61, opened 2021-09-28
fix install scheme for pypy3.8+

### LINKED ISSUE #56: pypy3.8 install schema has python3.8 in it

mattip:
Fix #56, continuation of #51

The install scheme in commands/install.py must agree with the scheme in `distutils/sysconfig.py`. It would be nice to refactor this to get the scheme from the stdlib `sysconfig.py` like in #23.

### DIFF
--- a/distutils/command/install.py
+++ b/distutils/command/install.py
@@ -19,6 +19,18 @@ from site import USER_BASE
 from site import USER_SITE
 HAS_USER_SITE = True
 
+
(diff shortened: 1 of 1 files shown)


mattip:
Ahh, #58 is better. Closing

PR #62, opened 2021-10-12
make GCC version regex more specific

GitMensch:
allowing also uncommon version identifiers to be matched correctly

Note: I've originally found a problem with the regex when working on https://github.com/msys2/MINGW-packages/pull/8353.
As this was half a year ago (the main issue that PR tackled is now solved differently so no need to fix that) I can't remember the details where the old one failed - please inspect the regex change and consider it when it works "as good as the old one" (I'm quite sure I've tested that back then).

Thank you!

### DIFF
--- a/distutils/cygwinccompiler.py
+++ b/distutils/cygwinccompiler.py
@@ -377,7 +377,7 @@ def check_config_h():
         return (CONFIG_H_UNCERTAIN,
                 "couldn't read '%s': %s" % (fn, exc.strerror))
 
-RE_VERSION = re.compile(br'(\d+\.\d+(\.\d+)*)')
+RE_VERSION = re.compile(br'[\D\s]*(\d+\.\d+(\.\d+)*)[\D\s]*$')
 
 def _find_exe_version(cmd):
     """Find the version of an executable by running `cmd` in the shell.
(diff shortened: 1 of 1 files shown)


lazka:
Imho all that version checking code could be removed. They check for 20(!!!) year old gcc/binutils

lazka:
Sadly numpy depends on all the internals, so not that easy :(

FFY00 (commented):
There's a failing test that is checking an old GCC version, it needs to be removed (only that bit of the test).

GitMensch:
For now I've just adjusted the regex again, now all of those tests samples pass. The real culprit is that I don't have my notes from several months ago which version string was broken... but with the optional spacing the new regex should match all those fine.

jaraco (changes requested):
> please inspect the regex change and consider it when it works "as good as the old one"

There are an infinite number of regexes that will work as good as the old one, but only a few are minimally satisfactory. Since all of the tests pass without this change, and especially because the problem is unknown, I'm disinclined to accept it. At the very least, I'd like to see a test that passes something to the regex that fails before this change and passes after the change along with a comment explaining under what conditions that version would be encountered. Otherwise, these extra characters could get optimized away and tests would continue to pass, and no one would have any reason to reject that optimization.

lazka:
> Imho all that version checking code could be removed. They check for 20(!!!) year old gcc/binutils

This is now happening in #77

PR #65, opened 2021-11-07
CCompiler.has_function: Do not fail if self.outputdir is set

### LINKED ISSUE #64: `has_function` broken when `CCompiler.output_dir` is set.
This has broken `has_function` when `CCompiler.output_dir` is set.

_Originally posted by @mkoeppe in https://github.com/pypa/distutils/issues/31#issuecomment-962520921_

mkoeppe:
Previously submitted as https://github.com/pypa/setuptools/pull/2860

Fixes #64

### DIFF
--- a/distutils/ccompiler.py
+++ b/distutils/ccompiler.py
@@ -802,7 +802,10 @@ int main (int argc, char **argv) {
         except (LinkError, TypeError):
             return False
         else:
-            os.remove("a.out")
+            output_dir = self.output_dir
+            if output_dir is None:
+                output_dir = ''
+            os.remove(os.path.join(output_dir, "a.out"))
         finally:
             for fn in objects:
                 os.remove(fn)
--- a/distutils/ccompiler.py
+++ b/distutils/ccompiler.py
@@ -802,10 +802,7 @@ int main (int argc, char **argv) {
         except (LinkError, TypeError):
             return False
         else:
-            output_dir = self.output_dir
-            if output_dir is None:
-                output_dir = ''
-            os.remove(os.path.join(output_dir, "a.out"))
+            os.remove(os.path.join(self.output_dir or '', "a.out"))
         finally:
             for fn in objects:
                 os.remove(fn)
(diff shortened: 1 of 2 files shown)


FFY00 (commented):
Aren't the other `os.remove` calls also affected?

mkoeppe:
> Aren't the other `os.remove` calls also affected?

No, they are using the result of `self.compile`

distutils/ccompiler.py
```diff
@@ -802,7 +802,10 @@ def has_function(self, funcname, includes=None, include_dirs=None,
         except (LinkError, TypeError):
             return False
         else:
-            os.remove("a.out")
+            output_dir = self.output_dir
+            if output_dir is None:
+                output_dir = ''
+            os.remove(os.path.join(output_dir, "a.out"))
```
FFY00, line 808:
```suggestion
            os.remove(os.path.join(self.output_dir or '', "a.out"))
```

FFY00 (approved):
Alright then. This looks okay to me, but I would like @jaraco to have a look, as I am not too familiar with this part of the code and don't have time to do an in-depth dive.

distutils/ccompiler.py
mkoeppe: My code follows the handling of `None` in other places in the file.

FFY00: Yeah, this is just a suggestion. It is purely stylistic. I'll leave it up to Jason.

jaraco: Both good suggestions. I like FFY00's suggestion, even though it's inconsistent. Since distutils is old code, I'll err on the side of using better syntax where possible.

mkoeppe:
Thanks.

jaraco:
I've discovered this test creates a directory 'scratch' with subdirectories:

```
distutils main $ find scratch
scratch
scratch/var
scratch/var/folders
scratch/var/folders/c6
scratch/var/folders/c6/v7hnmq453xb6p2dbz1gqc6rr0000gn
scratch/var/folders/c6/v7hnmq453xb6p2dbz1gqc6rr0000gn/T
```

Probably the test should clean this up.

PR #68, opened 2021-11-13
Add hook to enable monkeypatching in Debian and unblock distutils adoption in Setuptools.

### LINKED ISSUE #2: Install locations and features missed on Debian

jaraco:
- Create hooks and implement Debian's patches against those hooks. Fixes #2.
- Scheme now uses 'posix'
- Add a hook on import of distutils that will allow Debian to override the default behavior. Fixes #2.


jaraco:
Part of this PR includes https://github.com/pypa/distutils/blob/115da201da0e648d003dabbbd597d285d0d271f2/_distutils_system_mod.py. Debian maintainers can add this file somewhere to sys.path (plus augment install schemes in sysconfig) to allow this latest version of distutils to adopt the Debian patches as found in Python 3.9.

@doko42 Would you take a look and confirm this approach will unblock the use of this distutils in Setuptools?

This change may also help unblock issues like pypa/pip#10647.

cc @FFY00

jaraco:
It occurs to me that it may be easier to evaluate this approach by merging it and getting it into Setuptools such that a platform could experiment and test, so I'm going to proceed with the intention of iterating. I still welcome any feedback on the implementation.

jaraco:
In https://github.com/pypa/setuptools/issues/2956#issuecomment-1000895920 (https://github.com/pypa/distutils/commit/69f85733543b9f6257e89c4f79f9507f20d06b49), I extended the _distutils_system_mod to include an additional change to support Debian environments.

PR #70, opened 2021-11-19
Incorporate Fedora's distutil patch

hroncok:
See https://github.com/pypa/setuptools/pull/2896#issuecomment-973983395

### DIFF
--- a/distutils/command/install.py
+++ b/distutils/command/install.py
@@ -190,8 +190,6 @@ class install(Command):
 
     negative_opt = {'no-compile' : 'compile'}
 
-    # Allow Fedora to add components to the prefix
-    _prefix_addition = getattr(sysconfig, '_prefix_addition', "")
 
     def initialize_options(self):
         """Initializes options."""
@@ -473,10 +471,13 @@ class install(Command):
                     raise DistutilsOptionError(
                           "must not supply exec-prefix without prefix")
 
+                # Allow Fedora to add components to the prefix
+                _prefix_addition = getattr(sysconfig, '_prefix_addition', "")
+
                 self.prefix = (
-                    os.path.normpath(sys.prefix) + self._prefix_addition)
+                    os.path.normpath(sys.prefix) + _prefix_addition)
                 self.exec_prefix = (
-                    os.path.normpath(sys.exec_prefix) + self._prefix_addition)
+                    os.path.normpath(sys.exec_prefix) + _prefix_addition)
 
             else:
                 if self.exec_prefix is None:
--- a/distutils/command/install.py
+++ b/distutils/command/install.py
@@ -190,6 +190,8 @@ class install(Command):
 
     negative_opt = {'no-compile' : 'compile'}
 
+    # Allow Fedora to add components to the prefix
+    _prefix_addition = ""
 
     def initialize_options(self):
         """Initializes options."""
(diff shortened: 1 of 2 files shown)


hroncok:
This is a proof of concept. It has not been tested and Fedra's old Pythons currently don't set this attribute.

jaraco:
Thanks for this. For Debian and pkgsrc, I've been instead of incorporating the patches, creating hooks where platforms can provide their customizations. Let me take this example and illustrate what I have in mind.

jaraco:
With this latest change, Fedora would include on the Python path a module, `_distutils_system_mod` containing:

```python
"""
Patch distutils to mangle the prefix when not a virtualenv.

Sets self.prefix to sys.prefix + local/ unless RPM build.
"""

import os
import sysconfig
import distutils.command.install


rpm_build = 'RPM_BUILD_ROOT' in os.environ
if sysconfig._distutils_mangle_rpm_prefix and not rpm_build:
    distutils.command.install.install._prefix_addition = "/local"
```

Because the module is only available in the system environment, it doesn't need to do virtualenv detection (I hope). And because it's only available on platforms that introduce it, the `_distutils_mangle_rpm_prefix` can be known to be present (also an assumption).

Still, I think the general concept holds, and it gives Fedora more ownership over the behavior.

WDYT?

hroncok:
> Because the module is only available in the system environment

How so? Where would we put it if not to the stdlib which is always available? Even if we put it to site-packages, it will be available in virtual environments configured to see them.

OTOH _distutils_mangle_rpm_prefix does not need to exist at all, it was introduced here as a way to let distutils know we want to do that. With this hook, we are already doing that.

jaraco:
> > Because the module is only available in the system environment
> 
> How so? Where would we put it if not to the stdlib which is always available? Even if we put it to site-packages, it will be available in virtual environments configured to see them.

That's fine. I'd welcome any number of gates (restore virtualenv check) as needed if it's not possible to place the hook in such a context that it's only present when needed.

> OTOH _distutils_mangle_rpm_prefix does not need to exist at all, it was introduced here as a way to let distutils know we want to do that. With this hook, we are already doing that.

Sounds good. And of course, Fedora has full control over the behavior.

distutils/command/install.py
```diff
@@ -190,6 +190,8 @@ class install(Command):
 
     negative_opt = {'no-compile' : 'compile'}
 
+    # Allow Fedora to add components to the prefix
+    _prefix_addition = ""
```
jaraco, line 194: I'd also consider something like:

```suggestion
    _prefix_addition = getattr(sysconfig, '_prefix_addition', "")
```

And bypass the need for `_distutils_system_mod`.

hroncok: I wonder whether it would need to be a function, or if a constant works. I guess a constant *should* work, but I would need to double check.

jaraco: I guess it comes down to - is this value something that is static or determinable at startup time, or if it might need to change at run time based on other factors. If you're happy enough with the draft implementation, I can get that into a Setuptools release and you can experiment with a live release (using SETUPTOOLS_USE_DISTUTILS=local).

hroncok: Don't put anything into the release just yet, please. I can test this even without a release, I am just not as fast as you expect me to be (I do other things as well).

jaraco: Happy to oblige. I'm not trying to rush you; only to help. Thanks.

hroncok: Thanks. I expect to do some testing this week.

hroncok:
I've included this in https://github.com/hroncok/setuptools/tree/local-distutils-fedora-addition for easier testing.

hroncok:
This seems to work reliably.

jaraco:
Sounds good. Since you're not using `_distutils_system_mod`, I've removed the change to the comment authorizing that use. Is there anything else you want to see before rolling out this change?

distutils/command/install.py
```diff
@@ -190,6 +190,8 @@ class install(Command):
 
     negative_opt = {'no-compile' : 'compile'}
 
+    # Allow Fedora to add components to the prefix
+    _prefix_addition = getattr(sysconfig, '_prefix_addition', "")
```
jaraco, line 194: Since this is no longer patched globally, I'm tempted to move it inline with the other code rather than exposing a class attribute. WDYT?

jaraco: Done in 902e204.

jaraco:
I'm going to go ahead and roll this out. We can iterate more if needed.

hroncok:
Thanks.

martinpitt:
This now leads to disgusting and non-obvious hacks like https://github.com/candlepin/subscription-manager/pull/2989/files

Are you really sure that the intention is that `python3 ./setup.py install --prefix /usr --root /` installs to /usr/local? There is no obvious way to fix this now. Shouldn't this be limited to `pip install` somehow?

martinpitt:
Thanks @hroncok for the details! I'm mostly just the messenger here. Indeed I think it'd be better for https://github.com/candlepin/subscription-manager/ to actually install into /usr/local, and I'm discussing that in the mentioned PR. But still, it feels wrong that `--prefix=/usr` does that.

This happened with Fedora 36; ≤ 35 is fine.

hroncok:
It indeed feels wrong and we have a bugzilla open for that with details. I've liked it in some other place (I'm from mobile now, so harder for me to look it up again). It's not caused by any change here at pypa/distutils.

hroncok:
Here: https://bugzilla.redhat.com/show_bug.cgi?id=2026979

PR #71, opened 2021-11-21
Update `distutils.core.setup/run_setup` to better match `setuptools.build_meta`.

abravalheri:
## Motivation

I noticed that [`setuptools.build_meta`](https://github.com/pypa/setuptools/blob/413357393b74acd7f8bd02075a0f7ce8c239b869/setuptools/build_meta.py#L149-L158) "re-implements" the `run_setup` function (there are differences between the 2 functions, e.g. `distutils.core.run_setup` wraps the code in a `try...except`, but the main idea behind both look similar to me).

There is one major difference: `setuptools` use `tokenize.open` and replaces `\r\n` with `\n`.
I went through setuptools commit history but did not find any comments/commit messages explaining exactly why, but I assume that it [improves encoding detection](https://docs.python.org/3/library/tokenize.html?highlight=tokenize#tokenize.open).

My intention with this PR is to update the `distutils` code using the same technique as used in `setuptools`, so eventually the same function could be used.

## Changes

- Use the same `tokenize.open` technique as employed by `setuptools.build_meta` in `distutils.core.run_setup`.
- Additionally, I refactored the origin `setup` function and split out `run_commands`, this way, one can use `dist = run_setup("setup.py", stop_after="config")` to get the configuration and later run `run_commands(dist)`.
- I also added some tests.

### Side note

In terms of implications for `setuptools`, I imagine that eventually we could build the "new-static-config-style" distribution objects directly from the `setup.cfg`/`pyproject.toml` and then run them with `run_commands`, instead of always creating a `setup.py` stub.

Similarly, when `setup.py` is found, `dist = run_setup("setup.py", stop_after="config")` could be used to obtain the `dist` object and then revert to the same flow as a `setup.cfg`/`pyproject.toml`-only project...

(Basically a change in the mental model, instead of making `setup.cfg`/`pyproject.toml` use the same flow as `setup.py` one, we can make a `setup.py` use the same flow as `setup.cfg`).

### DIFF
--- a/distutils/core.py
+++ b/distutils/core.py
@@ -8,6 +8,7 @@ really defined in distutils.dist and distutils.cmd.
 
 import os
 import sys
+import tokenize
 
 from distutils.debug import DEBUG
 from distutils.errors import *
@@ -205,14 +206,17 @@ def run_setup (script_name, script_args=None, stop_after="run"):
     _setup_stop_after = stop_after
 
     save_argv = sys.argv.copy()
-    g = {'__file__': script_name}
+    g = {'__file__': script_name, '__name__': '__main__'}
     try:
         try:
             sys.argv[0] = script_name
             if script_args is not None:
                 sys.argv[1:] = script_args
-            with open(script_name, 'rb') as f:
-                exec(f.read(), g)
+            _open = getattr(tokenize, 'open', open)
+            # ^-- tokenize.open supports automatic encoding detection
+            with _open(script_name) as f:
+                code = f.read().replace(r'\r\n', r'\n')
+                exec(code, g)
         finally:
             sys.argv = save_argv
             _setup_stop_after = None
--- a/distutils/core.py
+++ b/distutils/core.py
@@ -224,9 +224,8 @@ def run_setup (script_name, script_args=None, stop_after="run"):
             sys.argv[0] = script_name
             if script_args is not None:
                 sys.argv[1:] = script_args
-            _open = getattr(tokenize, 'open', open)
-            # ^-- tokenize.open supports automatic encoding detection
-            with _open(script_name) as f:
+            # tokenize.open supports automatic encoding detection
+            with tokenize.open(script_name) as f:
                 code = f.read().replace(r'\r\n', r'\n')
                 exec(code, g)
         finally:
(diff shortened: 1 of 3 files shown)


distutils/core.py
```diff
@@ -205,14 +216,17 @@ def run_setup (script_name, script_args=None, stop_after="run"):
...
-    g = {'__file__': script_name}
+    g = {'__file__': script_name, '__name__': '__main__'}
     try:
         try:
             sys.argv[0] = script_name
             if script_args is not None:
                 sys.argv[1:] = script_args
-            with open(script_name, 'rb') as f:
-                exec(f.read(), g)
+            _open = getattr(tokenize, 'open', open)
```
abravalheri, line 225: If Python < 3.2 is not supported, this `getattr` could also be dropped.
I was not sure about which version `distutils` is targetting.

FFY00: It should be the same as setuptools, so this could be dropped.

FFY00 (approved):
This looks good to me, thanks! But I want @jaraco to have a look, just as a sanity check, before merging.

abravalheri:
Thank you very much @FFY00 for the review.

I updated he code to use `tokenize.open` directly, since it is available on Python >= 3.2

jaraco (approved):
This change is dangerously close to an enhancement. For context, I'd like to limit feature enhancements for distutils to only the most urgent (platform compatibility) until distutils is merged into Setuptools and maintained there. The reason is that the current state of affairs means that most users are using distutils from the stdlib and not from this repo. Until `SETUPTOOLS_USE_DISTUTILS=local` becomes the default, this code is largely un-validated in the wild. Limiting the divergence of this repo with stdlib distutils will limit the amount of divergence and thus surprise when this distutils becomes default.

That said, this change is primarily about aligning with Setuptools, whose technique _is_ proven and verified in the wild, so this change will actually aid in distutils adoption and limit the amount of divergence.

For that reason, let's proceed.

distutils/dist.py
```diff
@@ -168,6 +168,10 @@ def __init__(self, attrs=None):
         # for the setup script to override command classes
         self.cmdclass = {}
 
+        # Make sure 'commands' is defined, so dist.run_commands can run
+        # It might be overwritten by parse_command_line()
+        self.commands = []
+
```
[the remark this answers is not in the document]
jaraco: Removed in 405c150c1d.

abravalheri:
Thank you very much @jaraco. I will keep that in mind, and I totally understand what you explained.

Issue #34, opened 2021-03-12
Supporting MSYS2-MINGW python


naveen521kk:
[MSYS2](https://www.msys2.org/) provides Python which is compiled with GCC on Windows and is widely used in projects like [`pygobject`](https://pygobject.readthedocs.io) and anything related to Gtk, where compiling them for normal python is difficult. Their python works, I have used it many times, but I have seen a *lot* of patches to make things work. See a list [here](https://github.com/msys2/MINGW-packages/tree/master/mingw-w64-python). 

Currently, the following fails miserably because distutils always look for MSVC which isn't required in this version. Here, extensions should be compiled with GCC but support for that is missing in GCC, I think setuptools has them though, but doing something like this will fail.

```sh
$ export SETUPTOOLS_USE_DISTUTILS=local
$ python setup.py build
Traceback (most recent call last):
  File "setup.py", line 13, in <module>
    from setuptools import Extension, setup
  File "C:/msys64/mingw64/lib/python3.8/site-packages/setuptools/__init__.py", line 241, in <module>
    monkey.patch_all()
  File "C:/msys64/mingw64/lib/python3.8/site-packages/setuptools/monkey.py", line 99, in patch_all
    patch_for_msvc_specialized_compiler()
  File "C:/msys64/mingw64/lib/python3.8/site-packages/setuptools/monkey.py", line 162, in patch_for_msvc_specialized_compiler
    patch_func(*msvc9('find_vcvarsall'))
  File "C:/msys64/mingw64/lib/python3.8/site-packages/setuptools/monkey.py", line 149, in patch_params
    mod = import_module(mod_name)
  File "C:/msys64/mingw64/lib/python3.8/importlib/__init__.py", line 127, in import_module
    return _bootstrap._gcd_import(name[level:], package, level)
  File "C:/msys64/mingw64/lib/python3.8/site-packages/setuptools/_distutils/msvc9compiler.py", line 295, in <module>
    raise DistutilsPlatformError("VC %0.1f is not supported by this module" % VERSION)
distutils.errors.DistutilsPlatformError: VC 6.0 is not supported by this module
```

The authors can surely help on how to get those in.

Thanks.

lazka:
Just wanted to add that we are willing to help maintaining any cygwin/msys/mingw related parts in distutils if wanted. Just ping me if needed.

lazka:
A more general question:

[...]

PR #77, opened 2021-12-15
MinGW/Cygwin: Remove checks for ancient gcc/binutils

naveen521kk:
The versions checked here are 20 years old. Also dllwrap
has started to emit a deprecation warning in the latest release
spamming the build logs.

First steps towards https://github.com/pypa/distutils/issues/34
Commits from: https://github.com/msys2-contrib/cpython-mingw/commit/8ec77b61a8c33a8096882be9edfe0b23c4c62e2c and https://github.com/msys2-contrib/cpython-mingw/commit/825cae18785111ece6ac216335c4b83fccd6034c

cc @lazka

### DIFF
--- a/distutils/cygwinccompiler.py
+++ b/distutils/cygwinccompiler.py
@@ -51,13 +51,14 @@ import os
 import sys
 import copy
 import shlex
+import warnings
 from subprocess import check_output
 
 from distutils.unixccompiler import UnixCCompiler
 from distutils.file_util import write_file
 from distutils.errors import (DistutilsExecError, CCompilerError,
         CompileError, UnknownFileError)
-from distutils.version import LooseVersion
+from distutils.version import LooseVersion, suppress_known_deprecation
 
 def get_msvcr():
     """Include the appropriate MSVC runtime library if Python was built
@@ -122,12 +123,6 @@ class CygwinCCompiler(UnixCCompiler):
         self.cc = os.environ.get('CC', 'gcc')
         self.cxx = os.environ.get('CXX', 'g++')
 
-        # Older numpy dependend on this existing to check for ancient
-        # gcc versions. This doesn't make much sense with clang etc so
-        # just hardcode to something recent.
-        # https://github.com/numpy/numpy/pull/20333
-        self.gcc_version = LooseVersion("11.2.0")
-
         self.linker_dll = self.cc
         shared_option = "-shared"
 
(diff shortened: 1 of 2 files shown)


distutils/cygwinccompiler.py
```diff
@@ -124,33 +123,14 @@ def __init__(self, verbose=0, dry_run=0, force=0):
...
-            # dllwrap 2.10.90 is buggy
-            if self.ld_version >= "2.10.90":
-                self.linker_dll = self.cc
-            else:
-                self.linker_dll = "dllwrap"
+        # Older numpy dependend on this existing to check for ancient
+        # gcc versions. This doesn't make much sense with clang etc so
+        # just hardcode to something recent.
+        # https://github.com/numpy/numpy/pull/20333
+        self.gcc_version = LooseVersion("11.2.0")
```
lazka, line 130: I wonder if we should make this a property and emit a deprecation warning for this on access.

lazka, line 130: I wonder if we should make this a property and emit a deprecation warning on access.

distutils/cygwinccompiler.py
```diff
@@ -124,33 +123,14 @@ def __init__(self, verbose=0, dry_run=0, force=0):
         self.cc = os.environ.get('CC', 'gcc')
         self.cxx = os.environ.get('CXX', 'g++')
 
-        if ('gcc' in self.cc): # Start gcc workaround
-            self.gcc_version, self.ld_version, self.dllwrap_version = \
```
lazka, line 128: To clarify, this patch removes `self.ld_version` and `self.dllwrap_version` completely. If this is a potential problem re backwards compat (I personally doubt it) then we can try to set them to dummy values too.

distutils/cygwinccompiler.py
```diff
@@ -50,8 +50,7 @@
 import os
 import sys
 import copy
-from subprocess import Popen, PIPE, check_output
-import re
+from subprocess import Popen, check_output
 
 import distutils.version
 from distutils.unixccompiler import UnixCCompiler
```
lazka, line 56: Some of the imports are unused and could be removed. But they are used in the tests, so that would need fixing.

I initially kept them here to make the diff smaller.

naveen521kk: I noticed that. I didn't remove them because it was used in tests. I think it would be better if it was removed and the corresponding tests is changed.

distutils/cygwinccompiler.py
lazka: also we probably need to wrap it in `with distutils.version.suppress_known_deprecation():` after #75

distutils/cygwinccompiler.py
jaraco: Removed in 93e3790de8. It was indeed nicer to review with the smaller diff and then subsequently address this issue.

distutils/cygwinccompiler.py
jaraco: >  If this is a potential problem re backwards compat (I personally doubt it)

I appreciate the sentiment. Let's try without backwards compat and then revise if needed.

distutils/cygwinccompiler.py
jaraco: Agreed, I'd like to deprecate this property if it's not providing the value that it's name suggests.

GitMensch:
What is the reason for this being a draft?

distutils/cygwinccompiler.py
naveen521kk: I have done it in https://github.com/pypa/distutils/pull/77/commits/4d1035f0d52fb6c07b8603781c4a81122c0f7930

jaraco (approved):
Looks good. Let's get it out and see how it goes.

GitMensch:
@lazka When can we expect to see that in MSYS2 builds? I'd try to find some time to test it with a non-GCC compiler that actually work with GCC options (where I _think_ distutils's regex check on the version number previously stumbled over).

lazka:
It already is in the stdlib version for some time now.

naveen521kk:
Thanks!

PR #78, opened 2021-12-15
Import patches from MSYS2-MinGW

naveen521kk:
Fixes https://github.com/pypa/distutils/issues/34

These are from https://github.com/msys2-contrib/cpython-mingw/commits/mingw-v3.9.7/Lib/distutils. I have modified some of them for better logic and removed some unnecessary stuff from them.

This is a draft and built on top of #77 

cc @lazka

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -58,6 +58,28 @@ jobs:
         run: |
           pytest -rs
 
+  test_msys2_mingw:
+    strategy:
+      matrix:
+        include:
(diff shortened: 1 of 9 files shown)


naveen521kk:
> Closing as this work appears to be abandoned.

Sorry, just got busy with other things. I'll ping you when I get time to work on these. One thing this PR depends on for the tests to pass is https://github.com/msys2-contrib/cpython-mingw/pull/82 which again became stale.

PR #79, opened 2021-12-17
Honor sysconfig.get_preferred_scheme when selecting install schemes.

### LINKED ISSUE #76: Support Homebrew patches
From Homebrew's side we completely override the prefix (`base` var). This is because under our system, the Python install directory and the install directory that should be used for user-installed packages should be separate (or user packages are lost on upgrades).

We currently do this via `distutils.cfg` planted in the standard library distutils which will of course not work anymore.

Going forward we're looking at the system that will hopefully be introduced in Python 3.11 to add additional install schemes to sysconfig.py:

- https://gist.github.com/FFY00/625f65681fbcd7fc039dd4d727bb2c2f
- https://bugs.python.org/issue43976

Notably, the idea is schemes should be _added_ rather than replaced. This is why `sysconfig.get_preferred_scheme(key)` was introduced in Python 3.10. This is also important when dynamic logic may be required. Ideally I reckon distutils should use this when selecting a scheme.

While the above will cover Python 3.10+, there'll still be a question about what we do for <= 3.9. `sysconfig._prefix_addition` is almost what we want, though I don't know if calculating a relative path and going up levels (`..`) is a good idea.

_Originally posted by @Bo98 in https://github.com/pypa/setuptools/issues/2896#issuecomment-992774977_

jaraco:
Fixes #76.

### DIFF
--- a/distutils/command/install.py
+++ b/distutils/command/install.py
@@ -81,6 +81,11 @@ if HAS_USER_SITE:
         'data'   : '{userbase}',
         }
 
+    INSTALL_SCHEMES['osx_framework_user'] = {
+        'headers':
+            '{userbase}/include/{implementation_lower}{py_version_short}{abiflags}/{dist_name}',
+    }
+
 # The keys to an installation scheme; if any new types of files are to be
 # installed, be sure to add an entry to every installation scheme above,
 # and to SCHEME_KEYS here.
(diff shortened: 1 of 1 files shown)


jaraco:
@Bo98 Would you take a look at this patch and possibly test it in your environment? I did find that tests failed on my macOS install before I added the `osx_framework_user.headers`.

distutils/command/install.py
```diff
@@ -515,9 +520,17 @@ def finalize_other(self):
                       "I don't know how to install stuff on '%s'" % os.name)
 
     def select_scheme(self, name):
+        os_name, sep, key = name.partition('_')
```
jaraco, line 523: I acknowledge this approach is a little janky. I'd prefer to change the signature to accept the key instead, but I want to maintain the existing interface in case some of the transitional implementations are overriding it.

jaraco:
I'd like to proceed with this change. I'll roll it out and we can iterate subsequently.

PR #81, opened 2021-12-18
msvc9compiler: Don't raise DistutilsPlatformError on import

lazka:
It gets raised when get_build_version() returns a too low version
or when it fails to detect a MSVC version, which for exaple is the
case when CPython is built with gcc/clang.

In theory this module isn't needed on non-MSVC platforms, but setuptools
imports it anyway when monkey patching, which in turn makes setuptools fail.

To work around this only check the version when initialising MSVCCompiler().

This also mirrors the behaviour of the newer MSVCCompiler() (in _msvccompiler)
which also raises DistutilsPlatformError() only on initialisation in case
it can't find the right MSVC.

Part of #34

### DIFF
--- a/distutils/msvc9compiler.py
+++ b/distutils/msvc9compiler.py
@@ -291,8 +291,6 @@ def query_vcvarsall(version, arch="x86"):
 
 # More globals
 VERSION = get_build_version()
-if VERSION < 8.0:
-    raise DistutilsPlatformError("VC %0.1f is not supported by this module" % VERSION)
 # MACROS = MacroExpander(VERSION)
 
 class MSVCCompiler(CCompiler) :
@@ -339,6 +337,8 @@ class MSVCCompiler(CCompiler) :
     def initialize(self, plat_name=None):
         # multi-init means we would need to check platform same each time...
         assert not self.initialized, "don't init multiple times"
+        if self.__version < 8.0:
+            raise DistutilsPlatformError("VC %0.1f is not supported by this module" % self.__version)
         if plat_name is None:
             plat_name = get_platform()
         # sanity check for platforms to prevent obscure errors later.


lazka:
For context, here is the error on monkey patching:

```
$ export SETUPTOOLS_USE_DISTUTILS=local
$ python setup.py build
  File "setup.py", line 13, in <module>
    from setuptools import Extension, setup
  File "C:/msys64/mingw64/lib/python3.8/site-packages/setuptools/__init__.py", line 241, in <module>
    monkey.patch_all()
  File "C:/msys64/mingw64/lib/python3.8/site-packages/setuptools/monkey.py", line 99, in patch_all
    patch_for_msvc_specialized_compiler()
  File "C:/msys64/mingw64/lib/python3.8/site-packages/setuptools/monkey.py", line 162, in patch_for_msvc_specialized_compiler
    patch_func(*msvc9('find_vcvarsall'))
  File "C:/msys64/mingw64/lib/python3.8/site-packages/setuptools/monkey.py", line 149, in patch_params
    mod = import_module(mod_name)
  File "C:/msys64/mingw64/lib/python3.8/importlib/__init__.py", line 127, in import_module
    return _bootstrap._gcd_import(name[level:], package, level)
  File "C:/msys64/mingw64/lib/python3.8/site-packages/setuptools/_distutils/msvc9compiler.py", line 295, in <module>
    raise DistutilsPlatformError("VC %0.1f is not supported by this module" % VERSION)
distutils.errors.DistutilsPlatformError: VC 6.0 is not supported by this module
```

jaraco:
> This also mirrors the behaviour of the newer MSVCCompiler()

Thanks for tracing this detail, which makes it easy to justify the change.

jaraco:
I wonder if this change means the module can also be collected on non-windows systems.

PR #82, opened 2021-12-18
CI: Add a test job for Cygwin

lazka:
I'm not a Cygwin developer, but since the MSYS2 patches are touching Cygwin code paths this could be helpful in preventing regressions in that area.

Example run on my fork: https://github.com/lazka/distutils/runs/4569103775?check_suite_focus=true

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -21,6 +21,30 @@ jobs:
       - name: Run tests
         run: tox
 
+  test_cygwin:
+    strategy:
+      matrix:
+        python: [39]
+        platform: [windows-latest]
+    runs-on: ${{ matrix.platform }}
+    steps:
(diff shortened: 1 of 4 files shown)


jaraco:
This looks great. I wanted to take the opportunity here to notice that there was some redundancy across the two identified test modules and to consolidate that behavior, so I've done that in 1d98a15. It's entirely an iteration on your work and not a criticism.

lazka:
A bit of context: I'm not quite sure why the test try to set root uid/gid... maybe just because they are available and not the current user? I tried something similar by using the "Adminstrator" account and "Administrators" group on Windows first, which Windows is supposed to have by default, bot for some reason the GHA image is missing the former.

lazka:
Thanks!

PR #84, opened 2021-12-21
bpo-41282: Load install schemes from sysconfig (port from CPython)

henryiii:
In response to https://github.com/pypa/setuptools/issues/2946.

Directly applying the changes from https://github.com/python/cpython/pull/24549. No other testing at the moment.

### DIFF
--- a/distutils/command/install.py
+++ b/distutils/command/install.py
@@ -7,6 +7,7 @@ import os
 import contextlib
 import sysconfig
 import itertools
+import re
 
 from distutils import log
 from distutils.core import Command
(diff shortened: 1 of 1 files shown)


henryiii:
Okay, guessing we will need pypy and pypy_nt added (@mattip, correct?). Also, "headers" is not present, but why isn't it triggering on "include"? That error (on 3.6) is weird. @frenzymadness, perhaps?

Also, shouldn't any PR to distutils in CPython automatically be requested to make a PR here?

mattip:
Yes, PyPy 3.7 still needs `pypy` and `pypy_nt`. PyPy 3.8+ no longer does, we changed to the cpython layout schema (still using `implementation` and `implementation_lower` though).

jaraco:
> Also, shouldn't any PR to distutils in CPython automatically be requested to make a PR here?

Yes, probably. I wasn't fully grasping the implications of the changes to CPython when I reviewed it. My (incorrect) assessment was that the changes to stdlib were primarily removing functionality toward deprecation and that this project could adopt those changes as needed/desired. I agree, changes to CPython should be ported here.

jaraco:
Now I worry I may really have confused things with my analysis in pypa/setuptools#2946.

A few things:

- distutils has already adopted a _different_ technique for giving sysconfig primacy for install schemes (#23 and others).
- In that effort, I've identified that `headers` and `include` are not compatible, which is why `distutils` here still uses include.
- The PR that's most relevant to the issue is python/cpython#23142. That's the one where `_init_posix` was removed.

jaraco:
Apologies for the bad advice on the relevant PR, but distutils has diverged from cpython/distutils in this effort, so I'm going to close.

PR #87, opened 2021-12-22
Fix AttributeError when sysconfig does not supply platsubdir

### LINKED ISSUE #86: --user installs broken with setuptools 60.0.4
see reproduction in https://github.com/pypa/distutils/issues/85#issuecomment-999916730

broken by this patch: https://github.com/pypa/distutils/commit/eba2bcd310caa49b51aa640055e44ef8c74b5581

asottile:
resolves #86

### DIFF
--- a/distutils/command/install.py
+++ b/distutils/command/install.py
@@ -395,7 +395,7 @@ class install(Command):
                             'platlibdir': getattr(sys, 'platlibdir', 'lib'),
                             'implementation_lower': _get_implementation().lower(),
                             'implementation': _get_implementation(),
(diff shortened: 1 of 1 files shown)


FFY00 (approved):
Thank you for the fix. Unfortunately, almost everyone is unaware that most sysconfig variables are only available on posix-like operating systems, causing bugs like this to easily creep in.

hswong3i:
Confirm after applying this patch for v60.0.4, OBS packaging for deb/rpm with multiple OS looks good, see https://build.opensuse.org/package/show/home:alvistack/pypa-setuptools-60.0.4

PR #91, opened 2021-12-26
sysconfig: use get_config_h_filename() from the stdlib

lazka:
distutils.sysconfig provides various functions already present in
the stdlib sysconfig module.

Reusing the stdlib versions has the advantage that it is in control of
the Python distributor/packager and distutils doesn't have to deal with
platform details.

This tries to start the transition using the simple get_config_h_filename()
for starters by forwarding calls to the stdlib in case Python is
considered to be installed.

### DIFF
--- a/distutils/sysconfig.py
+++ b/distutils/sysconfig.py
@@ -13,6 +13,7 @@ import _imp
 import os
 import re
 import sys
+import sysconfig
 
 from .errors import DistutilsPlatformError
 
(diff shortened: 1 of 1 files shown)


FFY00 (approved):
Looks alright, but I'd like @jaraco to have a look too.

jaraco:
Looks good to me. Thanks!

PR #92, opened 2021-12-26
sysconfig: use get_makefile_filename() from stdlib sysconfig

lazka:
Instead of guessing the filename just refer to the stdlib.

This also removes the "_makefile_tmpl" hook added in pypa#16, but

1) It was never implemented by the distro requesting it from what I can see:
   https://github.com/NetBSD/pkgsrc/blob/586097714897b1b4d4a9/devel/py-setuptools/Makefile#L28
2) The stdlib version should return a proper result as it is patched by the distro requesting the change:
   https://github.com/NetBSD/pkgsrc/blob/6efa5763ec447864a7d4/lang/python38/patches/patch-Lib_sysconfig.py

Also adds a small test checking that the file exists on Unix platforms

### DIFF
--- a/distutils/sysconfig.py
+++ b/distutils/sysconfig.py
@@ -281,25 +281,9 @@ def get_config_h_filename():
 
 
 
-# Allow this value to be patched by pkgsrc. Ref pypa/distutils#16.
-_makefile_tmpl = 'config-{python_ver}{build_flags}{multiarch}'
-
-
 def get_makefile_filename():
     """Return full pathname of installed Makefile from the Python build."""
-    if python_build:
-        return os.path.join(_sys_home or project_base, "Makefile")
-    lib_dir = get_python_lib(plat_specific=0, standard_lib=1)
-    multiarch = (
-        '-%s' % sys.implementation._multiarch
-        if hasattr(sys.implementation, '_multiarch') else ''
-    )
-    config_file = _makefile_tmpl.format(
-        python_ver=get_python_version(),
-        build_flags=build_flags,
-        multiarch=multiarch,
-    )
-    return os.path.join(lib_dir, config_file, 'Makefile')
+    return sysconfig.get_makefile_filename()
 
 
 def parse_config_h(fp, g=None):
--- a/distutils/tests/test_sysconfig.py
+++ b/distutils/tests/test_sysconfig.py
@@ -38,6 +38,12 @@ class SysconfigTestCase(support.EnvironGuard, unittest.TestCase):
         config_h = sysconfig.get_config_h_filename()
         self.assertTrue(os.path.isfile(config_h), config_h)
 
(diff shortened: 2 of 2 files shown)


FFY00:
@0-wiz-0 can you confirm this does not break anything for NetBSD?

jaraco:
I do see that pkgsrc already [patches sysconfig](https://github.com/NetBSD/pkgsrc/blob/trunk/lang/python38/patches/patch-Lib_sysconfig.py) for similar purposes, so I expect that deferring to sysconfig is the right thing to do.

distutils/sysconfig.py
```diff
@@ -280,25 +281,11 @@ def get_config_h_filename():
     return os.path.join(inc_dir, 'pyconfig.h')
 
 
-# Allow this value to be patched by pkgsrc. Ref pypa/distutils#16.
-_makefile_tmpl = 'config-{python_ver}{build_flags}{multiarch}'
-
-
 def get_makefile_filename():
     """Return full pathname of installed Makefile from the Python build."""
     if python_build:
```
jaraco, line 286: I also notice that the sysconfig version has a _PYTHON_BUILD with similar logic. Can the entire function be delegated to sysconfig?

lazka: I wanted to keep this low-impact. But yes, I agree that Python probably knows better where its files are when run uninstalled. I can make that change if wanted.

lazka:
> I do see that pkgsrc already [patches sysconfig](https://github.com/NetBSD/pkgsrc/blob/trunk/lang/python38/patches/patch-Lib_sysconfig.py) for similar purposes, so I expect that deferring to sysconfig is the right thing to do.

Ah, thanks. I've updated the commit message accordingly, linking to the upstream patch,

distutils/sysconfig.py
lazka: What if I add a CI job running the test suite with an uninstalled CPython? (this would require building CPython, ensurepip, install pytest, run tests)

jaraco: I don’t feel strongly about my suggestion. Happy to take baby steps too.

Regarding CI, I’d be on board if slightly reluctant due to the slowness and maintenance burden. At some point, all tests should be migrated to Setuptools, so I dread slightly added complication for that migration.

That said, better to have something useful to migrate than no protection at all. If you’re willing to contribute it, I’d accept it.

jaraco (approved):
I’m ready to merge this anytime.

distutils/sysconfig.py
lazka: OK, thanks. I went with delegating everything to sysconfig now.

I tried running the tests uninstalled, but there is one (unrelated) error I don't understand yet, so I'll keep this in mind but wont pursue for now.

lazka:
thanks!

PR #93, opened 2021-12-26
sysconfig: use parse_config_h() from stdlib sysconfig

lazka:
The only difference is the argument name change


FFY00 (approved):
Looks good, thanks.

PR #95, opened 2021-12-27
tests: fix usage of requires_zlib() decorator

lazka:
It was passing the test function itself as an argument, skipping the test always.

See https://github.com/python/cpython/pull/28305 for the same fix in CPython.


lazka:
Ah, the API was changed in https://github.com/python/cpython/pull/20131

lazka:
Thanks!

PR #102, opened 2021-12-31
util.get_host_platform: use sysconfig.get_platform

naveen521kk:
sysconfig module has a copy of this function in
`sysconfig.get_platform` and would be better to
use that directly. This should help distributors
to patch that function in a single place,
ie sysconfig.

Co-authored-by:  Christoph Reiter <reiter.christoph@gmail.com>
Signed-off-by: Naveen M K <naveen521kk@gmail.com>

PR #104, opened 2022-01-03
util.get_host_platform: use sysconfig.get_platform (v2)

lazka:
This continues #102 but tries to address the concerns raised in https://github.com/pypa/distutils/pull/102#issuecomment-1004259287, see the second commit for any differences.

(I didn't add a test for aix since mocking seemed overly complex :/)

### DIFF
--- a/distutils/tests/test_util.py
+++ b/distutils/tests/test_util.py
@@ -2,6 +2,7 @@
 import os
 import sys
 import unittest
+import sysconfig as stdlib_sysconfig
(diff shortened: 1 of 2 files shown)


lazka:
(btw, I'd totally understand if you delay any mingw targeted patches until the setuptools transition pains are dealt with)

jaraco (approved):
LGTM. Thanks for looking into this and helping with the consolidation.

PR #106, opened 2022-01-05
[maint] removed unneded exception inclusion

Giuppox:
After [previous commit](https://github.com/pypa/distutils/commit/221e8715518c23a45b5b81d884d05035e9c896dd), `DistutilsPlatformError` inclusion is not needed anymore since macos version's check has been moved to `util.py`.


jaraco:
Thanks!

PR #113, opened 2022-01-24
Fixes #112 install command doesn't use platform in nt_user scheme

### LINKED ISSUE #112: install command doesn't use platform in user site directory on Windows
In `sysconfig`, the `nt_user` scheme includes a platform specifier to distinguish between 32-bit, 64-bit and ARM64 builds.

This is missing from the `install` command, which is used by `setuptools` (and indirectly, pip) to determine the location of the user site in some cases. As a result, on non-x64 platforms, the path is incorrect.

zooba:
Fixes #112

### DIFF
--- a/distutils/command/install.py
+++ b/distutils/command/install.py
@@ -411,17 +411,19 @@ class install(Command):
                             'implementation_lower': _get_implementation().lower(),
                             'implementation': _get_implementation(),
                            }
-        try:
-            local_vars['py_version_nodot_plat'] = sys.winver.replace('.', '')
-        except AttributeError:
-            local_vars['py_version_nodot_plat'] = ''
+
+        # vars for compatibility on older Pythons
+        compat_vars = dict(
+            # Python 3.9 and earlier
+            py_version_nodot_plat=getattr(sys, 'winver', '').replace('.', ''),
+        )
 
         if HAS_USER_SITE:
             local_vars['userbase'] = self.install_userbase
             local_vars['usersite'] = self.install_usersite
 
         self.config_vars = _collections.DictStack(
-            [sysconfig.get_config_vars(), local_vars])
+            [compat_vars, sysconfig.get_config_vars(), local_vars])
 
         self.expand_basedirs()
 
--- a/distutils/command/install.py
+++ b/distutils/command/install.py
@@ -68,8 +68,8 @@ if HAS_USER_SITE:
     INSTALL_SCHEMES['nt_user'] = {
         'purelib': '{usersite}',
         'platlib': '{usersite}',
-        'headers': '{userbase}/{implementation}{py_version_nodot}/Include/{dist_name}',
-        'scripts': '{userbase}/{implementation}{py_version_nodot}/Scripts',
+        'headers': '{userbase}/{implementation}{py_version_nodot_plat}/Include/{dist_name}',
+        'scripts': '{userbase}/{implementation}{py_version_nodot_plat}/Scripts',
         'data'   : '{userbase}',
         }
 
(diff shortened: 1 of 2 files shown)


zooba:
Still needs a specific test, and there are issues with the existing test that need fixing, so please do not merge until I've done those

zooba:
I think this is ready to commit. Bit weird that distutils.sysconfig and sysconfig disagree[d?] on the include directory, but it seems to be handled okay.

distutils/command/install.py
```diff
@@ -68,8 +68,8 @@
     INSTALL_SCHEMES['nt_user'] = {
         'purelib': '{usersite}',
         'platlib': '{usersite}',
-        'headers': '{userbase}/{implementation}{py_version_nodot}/Include/{dist_name}',
-        'scripts': '{userbase}/{implementation}{py_version_nodot}/Scripts',
+        'headers': '{userbase}/{implementation}{py_version_nodot_plat}/Include/{dist_name}',
+        'scripts': '{userbase}/{implementation}{py_version_nodot_plat}/Scripts',
```
jaraco, line 72: Are these schemes already updated in `sysconfig` for Python 3.7+? If so, perhaps the best course of action would be to remove this scheme altogether and rely on the definition as found in `sysconfig`.

distutils/command/install.py
```diff
@@ -411,6 +411,10 @@ def finalize_options(self):
                             'implementation_lower': _get_implementation().lower(),
                             'implementation': _get_implementation(),
                            }
+        try:
+            local_vars['py_version_nodot_plat'] = sys.winver.replace('.', '')
+        except AttributeError:
+            local_vars['py_version_nodot_plat'] = ''
```
jaraco, line 417: Same here - is this value reliably set in sysconfig.get_config_vars()? If so, it should be possible to omit this work.

distutils/tests/test_install.py
```diff
@@ -122,6 +124,18 @@ def cleanup():
...
 
+        actual_headers = os.path.relpath(cmd.install_headers, self.user_base)
+        if os.name == 'nt':
+            expect_headers = os.path.join(
+                os.path.relpath(os.path.dirname(self.old_user_site), self.old_user_base),
+                'Include',
+                'xx',
+            )
+        else:
+            expect_headers = os.path.join(sysconfig.get_python_inc(0, ''), 'xx')
```
jaraco, line 135: I'm struggling to understand the intention behind divergent tests here. It's annoying that the tests need to have such intimate knowledge of the implementation details in order to set expectations. Is there a better way that derives this expectation from the API?

Is it possible the expected value could be derived from some value in `sysconfig` (not `distutils.sysconfig`)? I'm trying to move more of the logic away from `distutils.sysconfig` as that module is deprecated, so this line moves in the wrong direction.

distutils/command/install.py
jaraco: Alright. Nevermind. I see now how these are needed, because py_version_nodot_plat was only added in Python 3.10.

distutils/command/install.py
jaraco: This value was only added in Python 3.10, so is needed at least for older Pythons.

distutils/tests/test_install.py
zooba: It's because Windows doesn't include the runtime version in its headers subdirectory (`{prefix}\Include\xx`), while POSIX does (`{prefix}\include\python3.10\xx`).

But POSIX also seems to include the ABI flags in the name? Sometimes? I'm sure the logic is "fine", but I couldn't figure it out. Pretty sure it's also changed upstream with rebasing onto `sysconfig`, though not in any way that impacts default builds after 3.7.

If you weren't running 3.7 tests on this repo, I wouldn't even have noticed the change and would've stuck with my simpler test logic :)

Issue #124, opened 2022-03-04
[BUG] build_scripts.py incorrectly handles shebang lines with DOS line endings


amalachow:
Duplicating https://github.com/pypa/setuptools/issues/3129 here due to some permissions issues during the transfer.

### Python version

Python 3.8.10

### OS

NT-10.0-19042

### Description

shebang using DOS line endings causes the shebang regular expression to fail resulting in redirection to the python interpreter used for the installed script to not behave as expected.

### Expected behavior

The shebang is rewritten consistently for UNIX and DOS line endings.

### How to Reproduce

```bash
#!/bin/bash -eu
printf "[[System Info]]\n"
uname -srvmo
python3 --version
python3 -m virtualenv --version
rm -rf .venv
python3 -m virtualenv --quiet -p python3 .venv
.venv/Scripts/python -m pip install --quiet --upgrade pip &> /dev/null
.venv/Scripts/python -m pip install --quiet --upgrade setuptools &> /dev/null
.venv/Scripts/python -m pip list

rm -rf example
mkdir example
cat > example/example.py << END
#!/usr/bin/env python3

def main():
    print("Example script")

if __name__=="__main__":
    main()
END

cat > example/setup.py << END
from setuptools import setup
setup(name="example", scripts=["example.py"])
END

printf "\n[[Expected behavior]]\n"
file example/example.py
.venv/Scripts/pip install --quiet ./example
head -1 .venv/Scripts/example.py

printf "\n[[Issue]]\n"
sed -ie 's/\r*$/\r/g' example/example.py
file example/example.py
.venv/Scripts/pip install --quiet -I ./example
head -1 .venv/Scripts/example.py

fix="--- a/build_scripts.py
+++ b/build_scripts.py
@@ -15 +15 @@
-first_line_re = re.compile(b'^#!.*python[0-9.]*([ \t].*)?$')
+first_line_re = re.compile(b'^#!.*python[0-9.]*([ \t].*)?\r?$')"

printf "\n[[Proposed Fix]]\n"
echo "${fix}"
patch -u -p0 .venv/Lib/site-packages/setuptools/_distutils/command/build_scripts.py <<< ${fix}
file example/example.py
.venv/Scripts/pip install --quiet -I ./example
head -1 .venv/Scripts/example.py```
```

### Output

```console
[[System Info]]
MINGW64_NT-10.0-19042 3.1.7-340.x86_64 2021-03-26 22:17 UTC x86_64 Msys
Python 3.8.10
[...]

PR #125, opened 2022-03-04
Support DOS EOL in build_scripts shebang RE

amalachow:
Addresses #124


jaraco:
Per discussion in the issue, I'd like to address this issue differently.

PR #133, opened 2022-04-09
Use cache_tag in default build_platlib dir

mgorny:
Use `sys.implementation.cache_tag` instead of Python version to create
the default directory for `build_platlib`.  This guarantees that
the directories used by CPython and PyPy are distinct.  Prior to
the change, both CPython and PyPy would use e.g. `lib.linux-x86_64-3.9`.
With the change, they are going to use `lib.linux-x86_64-cpython39`
and `lib.linux-x86_64-pypy39` respectively.

### DIFF
--- a/distutils/command/build.py
+++ b/distutils/command/build.py
@@ -81,7 +81,8 @@ class build(Command):
                             "--plat-name only supported on Windows (try "
                             "using './configure --help' on your platform)")
 
-        plat_specifier = ".%s-%d.%d" % (self.plat_name, *sys.version_info[:2])
+        plat_specifier = ".%s-%s" % (self.plat_name,
+                                     sys.implementation.cache_tag)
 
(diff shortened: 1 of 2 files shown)


kmod:
Hello, I'd like to advocate for a different name for these directories. There is a bit of a convention of using `sysconfig.get_config_var("VERSION")` for this purpose instead of `sys.implementation.cache_tag`, so these new cache directories will be named differently than most of the ecosystem expects.  For example
- [within distutils](https://github.com/pypa/distutils/blob/e28e150c036e8b2bd55a37df13e836e098b5f37d/distutils/command/build_scripts.py#L104-L105)
- [how uwsgi does it](https://github.com/unbit/uwsgi/blob/master/plugins/python/uwsgiplugin.py#L54)

These directory names are used for significantly more than caching, so I believe `VERSION` is a bit more appropriate. See #107 for a bit of rationale and my proposed code change.

PR #135, opened 2022-04-09
Update test_home_installation_scheme for pypy install paths

mgorny:


distutils/tests/test_install.py
```diff
@@ -56,14 +56,17 @@ def check_path(got, expected):
             expected = os.path.normpath(expected)
             self.assertEqual(got, expected)
 
-        libdir = os.path.join(destination, "lib", "python")
+        impl_name = sys.implementation.name
+        if impl_name == "cpython":
+            impl_name = "python"
```
jaraco, line 61: It's a shame this incongruity has to exist, but since it does, better to document it.

mgorny:
Thanks!

PR #137, opened 2022-04-21
Detect when a venv is created from an in-tree build. Fixes #132

### LINKED ISSUE #132: distutils.sysconfig can't handle venv from source build on Windows
If you build CPython from source, run it from the source tree to create a venv, then use that venv to build things, the `distutils.sysconfig` module never finds its way back to the original files. This is because some of its tests for "are we a source tree build" fail due to the venv.

This may apply on other platforms as well, but it's going to behave differently because of different on-disk layouts, so I'll stick to Windows.

The first of these is here:
https://github.com/pypa/distutils/blob/2a233e54eba61e6dd6fc0c78f7f844430a758498/distutils/sysconfig.py#L54-L61

This snippet also needs to test against `BASE_PREFIX` in case the incoming path is against `sys.base_prefix`.

```
         if d and os.path.normcase(d).startswith( 
                 os.path.normcase(os.path.join(BASE_PREFIX, "PCbuild"))): 
             return BASE_PREFIX 
         return d 
```

Fixing that should ensure this part of `get_python_inc` further down goes through the right path:
https://github.com/pypa/distutils/blob/2a233e54eba61e6dd6fc0c78f7f844430a758498/distutils/sysconfig.py#L124-L130

I haven't walked all the rest of the way through the code - I'm pretty sure somewhere there's code calculating `libs` incorrectly, but wanted to get this much reported so it isn't lost.

zooba:


zooba:
As I mentioned in #132 I suspect there is more to do to really make this work, but this is at least enough to unblock whatever the next issue happens to be.

PR #138, opened 2022-04-27
Optional fields are really optional

wimglenn:
- make all fields documented as optional in [the spec](https://packaging.python.org/en/latest/specifications/core-metadata/) truly optional (and don't spam warnings when they've been omitted)

- stop coercing missing values to the string "UNKNOWN"

_Note:_ the setuptools integration is expected to fail, because this PR needs to be coordinated with changes in `setuptools/dist.py` too.  I'll prepare a separate PR on pypa/setuptools for that part.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -11,9 +11,9 @@ jobs:
     strategy:
       matrix:
         python:
-        - 3.7
-        - 3.8
-        - 3.9
+        - "3.7"
+        - "3.8"
+        - "3.9"
         - "3.10"
         platform:
         - ubuntu-latest
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -11,9 +11,9 @@ jobs:
     strategy:
       matrix:
         python:
-        - "3.7"
-        - "3.8"
-        - "3.9"
+        - 3.7
+        - 3.8
+        - 3.9
         - "3.10"
         platform:
         - ubuntu-latest
--- a/distutils/dist.py
+++ b/distutils/dist.py
@@ -1064,7 +1064,7 @@ class DistributionMetadata:
 
         def _read_field(name):
             value = msg[name]
-            if value:
+            if value and value != "UNKNOWN":
                 return value
 
         def _read_list(name):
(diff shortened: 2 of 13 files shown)


distutils/dist.py
```diff
@@ -1125,20 +1124,32 @@ def write_pkg_file(self, file):
...
+        author_email = self.get_contact_email()
+        if author_email:
+            file.write('Author-email: %s\n' % author_email)
+        license = self.get_license()
+        if license:
+            file.write('License: %s\n' % license)
         if self.download_url:
             file.write('Download-URL: %s\n' % self.download_url)
-
-        long_desc = rfc822_escape(self.get_long_description())
```
abravalheri, line 1139: I have the impression that RFC 822 escaping is required by the core metadata spec: https://packaging.python.org/en/latest/specifications/core-metadata/#description

Maybe the escaping in distutils differs from the one in the spec, instead of 8 spaces, it should be 7 spaces and a |, but no escaping might confuse the libraries reading the file.

distutils/dist.py
```diff
@@ -1125,20 +1124,32 @@ def write_pkg_file(self, file):
...
-        file.write('Summary: %s\n' % self.get_description())
-        file.write('Home-page: %s\n' % self.get_url())
-        file.write('Author: %s\n' % self.get_contact())
-        file.write('Author-email: %s\n' % self.get_contact_email())
-        file.write('License: %s\n' % self.get_license())
+
+        # optional fields
+        summary = self.get_description()
+        if summary:
+            file.write('Summary: %s\n' % summary)
```
jaraco, line 1135: I'd like to see this pattern replaced by a single function call that fits on one line, something like `maybe_write(file, 'Summary', self.get_description())`

distutils/tests/test_install.py
```diff
@@ -209,16 +211,19 @@ def test_record(self):
...
     def test_record_extensions(self):
         cmd = test_support.missing_compiler_executable()
         if cmd is not None:
             self.skipTest('The %r command is not found' % cmd)
         install_dir = self.mkdtemp()
-        project_dir, dist = self.create_dist(ext_modules=[
-            Extension('xx', ['xxmodule.c'])])
+        project_dir, dist = self.create_dist(
+            name="testdist",
+            version="0.1",
```
jaraco, line 224: In other places, I've argued against requiring to supply name and version, that "UNKNOWN" and 0.0.0 are good defaults for a lot of cases, tests especially. Here's an example where the test is having to introduce extra lines and values irrelevant to the behavior under test, distracting from the essential behavior.

distutils/dist.py
```diff
@@ -1064,9 +1064,8 @@ def read_pkg_file(self, file):
 
         def _read_field(name):
             value = msg[name]
-            if value == 'UNKNOWN':
```
jaraco, line 1067: I'm pretty sure this is here to replace UNKNOWN when found in an existing metadata file. How can we be confident this behavior isn't important for existing metadata?

.github/workflows/main.yml
```diff
@@ -11,9 +11,9 @@ jobs:
     strategy:
       matrix:
         python:
-        - 3.7
-        - 3.8
-        - 3.9
+        - "3.7"
+        - "3.8"
+        - "3.9"
```
jaraco, line 16: This change seems irrelevant to the goal. Can you back it out?

jaraco:
Overall, this looks good to me. Is there any way this change could be rolled out in such a way that it doesn't have to be synchronized with Setuptools? Perhaps to have a PR land there first? I'm concerned the coordination with Setuptools may be indicative of a compatibility issue with stdlib distutils. I'll check the Setuptools PR soon.

wimglenn:
@jaraco  IIUC you want the changes in setuptools to work regardless of whether it calls to pypa/distutils or stdlib distutils?

jaraco:
> @jaraco IIUC you want the changes in setuptools to work regardless of whether it calls to pypa/distutils or stdlib distutils?

That's correct. Setuptools still supports using stdlib distutils, so any change there will need to support both. If this change requires Setuptools to drop support for stdlib distutils, the change will have to wait until that happens (maybe this year).

wimglenn:
@jaraco  It is not necessary to drop support for stdlib distutils. More details [here](https://github.com/pypa/setuptools/pull/3289#issuecomment-1114443248).

Issue #73, opened 2021-12-05
Support Cygwin patches


jaraco:
Also need to incorporate mingw/cygwin distutil patches - see https://github.com/pypa/distutils/pull/60#issuecomment-973277696, https://github.com/pypa/distutils/pull/60#issuecomment-980846735;
pypa/setuptools#2909 adds a CI for Cygwin

_Originally posted by @mkoeppe in https://github.com/pypa/setuptools/issues/2896#issuecomment-986278425_

jaraco:
@mkoeppe Do you have a contact for Cygwin? Do you know what the Cygwin patches are?

Setuptools released two versions, 48 and  50, which twice attempted to adopt distutils into Setuptools, but there were emergent issues. I attempted to capture those emergent issues, but did not see anything reported from cygwin.

I'd really like to get this feature rolled out, as the longer it takes to get distutils adoption as the default, the more risk there is that changes to distutils will break use-cases that aren't being exercised.

Cygwin users will have the option to opt out by using `SETUPTOOLS_USE_DISTUTILS=stdlib`, but I would prefer if users did not have to do that, but I'm also reluctant to delay the rollout further.

Please do let us know more about what patches are present and how distutils might be able to incorporate that behavior.

mkoeppe:
I don't have a cygwin contact, but the patches can be found at https://cygwin.com/git-cygwin-packages/?p=git/cygwin-packages/python38.git;a=tree

mkoeppe:
In particular https://cygwin.com/git-cygwin-packages/?p=git/cygwin-packages/python38.git;a=blob;f=3.1-enable-new-dtags.patch;h=716af3b14d3483e5a4cdb65d0ec45ba92bd03ceb;hb=HEAD corresponds to the issue that I ran into that I mentioned in https://github.com/pypa/distutils/pull/60#issuecomment-980846735

jaraco:
I've [reached out](https://cygwin.com/pipermail/cygwin/2021-December/250185.html) on the mailing list to raise awareness of the issue.

mkoeppe:
Same for MSYS2 (mingw). Patches are here: https://github.com/msys2/MINGW-packages/tree/master/mingw-w64-python

mkoeppe:
For testing on MSYS2, there is a GH Action: https://github.com/msys2/setup-msys2

jaraco:
Let’s track that project separately, even if some of the solutions may be similar.

jaraco:
[...]

PR #139, opened 2022-05-04
Test that find_library_file works with Cygwin libraries

DWesl:
It worked for setuptools<60 and doesn't with setuptools>=60; I would like it to keep working.

Follow-up to python-pillow/pillow#6216

This is one of the patches mentioned in https://github.com/pypa/distutils/issues/73#issuecomment-991923804

### DIFF
--- a/distutils/cygwinccompiler.py
+++ b/distutils/cygwinccompiler.py
@@ -101,9 +101,11 @@ class CygwinCCompiler(UnixCCompiler):
     compiler_type = 'cygwin'
     obj_extension = ".o"
     static_lib_extension = ".a"
-    shared_lib_extension = ".dll"
+    shared_lib_extension = ".dll.a"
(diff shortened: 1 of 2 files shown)


DWesl:
Cygwin test failing in the expected place

DWesl:
The new test passes with the library name pattern update.  Merging this should probably wait for input from a maintainer about what `CCompiler.find_library_file` is supposed to be looking for: should it find any file related to the given library, regardless of whether it can be used in linking, or only those files useful in linking?

The `cyg${name}.dll` files aren't used in linking the way the `lib${name}{,.dll}.a` files are, but it is the file needed by `dlopen`.

DWesl:
Worst case for the problem I'm thinking of involves the linker failing, and I think those error messages are decently informative, so that's a decent approach.

PR #140, opened 2022-05-04
Update Cygwin library name patterns

DWesl:
Shared and static libraries are under `/usr/lib/`; the dylibs are under `/usr/bin`; I think that's a sensible way to assign names to the three things (shared lib being import library, the thing needed at link time, with the dylib being needed at run time).

Replaces pypa/setuptools#3304; follow-up to python-pillow/pillow#6216.  Tests in #139.


DWesl:
Replaced by #139

PR #143, opened 2022-05-18
Fix _is_gcc() for cross-compiling and treat clang/clang++ as gcc/g++

zboszor:
This patch allows detecting gcc in the Yocto cross-compiler system better.
The compiler is not just called gcc or g++, the base name of the compiler may have a machine triplet prefix.
Treat clang and clang++ the same as gcc and g++ as they mimic the command line switches.

### DIFF
--- a/distutils/unixccompiler.py
+++ b/distutils/unixccompiler.py
@@ -260,7 +260,8 @@ class UnixCCompiler(CCompiler):
         return "-L" + dir
 
     def _is_gcc(self, compiler_name):
-        return "gcc" in compiler_name or "g++" in compiler_name
+        cnpat = re.compile('.*(gcc|g\+\+|clang|clang\+\+)$')
+        return not (cnpat.match(compiler_name) is None)
 
     def runtime_library_dir_option(self, dir):
         # XXX Hackish, at the very least.  See Python bug #445902:


rossburton:
Surely `'gcc' in compiler_name` works for a triplet-prefixed compiler name?

Also arguably clang is not gcc, so maybe there should be a `is_clang` function that can be use as `is_gcc() or is_clang()` for situations where either work.

zboszor:
The `in` is also a problem. Any command can have a "gcc" substring somewhere in the middle and not being compiler related.

It's important that the command name ends in the pattern.

Maybe compiler_name.endswith() would be a better choice but a regex was less typing and more compact then spelling out compiler_name.endswith() 4 times on the same line.

Also, the logic was used to decide passing `-Wl,-R` vs just `-R` and clang is equivalent with GCC in this regard.

zboszor:
> `_is_gcc` is used only for rpath setting on `hp-ux`. Are you try to cross compile for `hp-ux`?

Hm. In older versions it wasn't so.
The copy of distutils in Python and setuptools certainly uses an older version.
This PR was originally made for setuptools, then I was pointed here.
It's still a problem in older Yocto releases with old versions of python3-setuptools.

abravalheri:
> The copy of distutils in Python and setuptools certainly uses an older version.

@zboszor, if you install today the latest version of setuptools from PyPI, the bundled copy of `distutils` is pretty much similar to this repository, maybe just a few commits behind (they are synched frequently).

If you set `SETUPTOOLS_USE_DISTUTILS=stdlib` then the copy in Python's stdlib will be used (for which there is not much hope to add changes/patches, I am afraid...).

jaraco:
It sounds like from the discussion that this change is unneeded. If it is needed, please first file an issue describing in more detail what's not working with the current implementation. Also, a change like the one proposed should also include a test to capture the new expectation.

PR #144, opened 2022-05-18
Fixed Issue on Windows+MSYS2 with unsafe handling of CC config variable.

MilchRatchet:
On some platforms like Windows in combination with MSYS2 this part of code gets executed but the CC config variable does not exist.

I added a check for if the CC config variable was found and moved the code to the only platform case that requires it.


jaraco:
> I added a check for if the CC config variable was found and moved the code to the only platform case that requires it.

It seems to me you fixed the problem twice, once by putting a conditional on the CC var and again by moving the check outside the scope where it was missing, thereby creating new logic that's no longer relevant. Especially since this logic is not tested, let's limit the change to the minimum suitable. I'll remove the conditional on the CC var.

PR #161, opened 2022-07-31
DOC: Extend Windows rpath error message with likely cause

DWesl:
Extend the error message for attempting to call `runtime_library_dir_option` on Windows with the most likely way I have found to trigger the call, creating an `Extension` that specifies `runtime_library_dirs`.

Hopefully this will give users something to search for in their `setup.py` and a starting point for what to change.  Inspired by pypa/setuptools#3450.

Since this is now an error, should this be documented in [the setuptools documentation for `Extension`](https://setuptools.pypa.io/en/latest/userguide/ext_modules.html#extension-api-reference) as well?

### DIFF
--- a/distutils/cygwinccompiler.py
+++ b/distutils/cygwinccompiler.py
@@ -341,7 +341,9 @@ class Mingw32CCompiler(CygwinCCompiler):
 
     def runtime_library_dir_option(self, dir):
         raise DistutilsPlatformError(
-            "don't know how to set runtime library search path on Windows"
+            "don't know how to set runtime library search path on Windows\n"
+            "Most likely some Extension specifies 'runtime_library_dirs'\n"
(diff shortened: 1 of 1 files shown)


distutils/cygwinccompiler.py
```diff
@@ -341,7 +341,9 @@ def __init__(self, verbose=0, dry_run=0, force=0):
 
     def runtime_library_dir_option(self, dir):
         raise DistutilsPlatformError(
-            "don't know how to set runtime library search path on Windows"
+            "don't know how to set runtime library search path on Windows\n"
```
jaraco, line 344: The original message appears to be a copy/paste from line 279. Probably we'll want to consolidate the messaging (even though one is an error and one is a warning). I'll use this commit as inspiration for a more concise message.

PR #163, opened 2022-08-01
Fix finding library files

TimotheusBachinger:
Since setuptools 63.3.0, pillow couldn't be built anymore, see https://github.com/python-pillow/Pillow/issues/6471
This was broken after refactoring: https://github.com/pypa/setuptools/commit/25dcca2902780146a1ee298f18648334eb3cddff

I am now directly creating the PR here as discussed in https://github.com/pypa/setuptools/pull/3476

### DIFF
--- a/distutils/unixccompiler.py
+++ b/distutils/unixccompiler.py
@@ -373,8 +373,8 @@ class UnixCCompiler(CCompiler):
 
         searched = (
             os.path.join(root, lib_name)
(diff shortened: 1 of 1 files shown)


TimotheusBachinger:
closed as it's fixed now in https://github.com/pypa/distutils/commit/f700a6b5f9965b3a886827d523271a2c9388edb5

PR #173, opened 2022-08-21
Fix, again, finding headers during cross compiling

rnhmjoj:
Find the correct headers (from the host machine, not build) when cross compiling.
This should have been fixed in https://github.com/pypa/distutils/pull/145 but a follow up commit broke it again.

Note: I tested that normal and cross compilation of a package with native extensions succeed on a POSIX system.


jaraco:
Thanks for this and thanks for linking to the issue. It sure would be nice if the test suite contained either unit tests or integration tests that captured some of the cross-compiling requirements. As it is, the code is the spec, so it's difficult to safely refactor.

PR #184, opened 2022-10-08
Add support for building extensions using MinGW compilers

naveen521kk:
See individual commits for details on what this PR changes.

A new version of https://github.com/pypa/distutils/pull/78
Fixes https://github.com/pypa/distutils/issues/34

cc @lazka

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -115,6 +115,45 @@ jobs:
         shell: C:\cygwin\bin\env.exe CYGWIN_NOWINPATH=1 CHERE_INVOKING=1 C:\cygwin\bin\bash.exe -leo pipefail -o igncr {0}
         run: tox
 
+  test_msys2_mingw:
+    strategy:
+      matrix:
+        include:
+          - { sys: mingw64, env: x86_64 }
+          - { sys: mingw32, env: i686 }
+          - { sys: ucrt64,  env: ucrt-x86_64 }
+          - { sys: clang64, env: clang-x86_64 }
+    runs-on: windows-latest
+    steps:
+      - uses: actions/checkout@v4
+      - uses: msys2/setup-msys2@v2
+        with:
+          msystem: ${{matrix.sys}}
+          install: |
+            mingw-w64-${{matrix.env}}-toolchain
+            mingw-w64-${{matrix.env}}-python
+            mingw-w64-${{matrix.env}}-python-pip
+            mingw-w64-${{matrix.env}}-python-virtualenv
+            mingw-w64-${{matrix.env}}-cc
+            git
+      - name: Install Dependencies
+        shell: msys2 {0}
+        run: |
+          export VIRTUALENV_NO_SETUPTOOLS=1
+
+          python -m virtualenv venv
+          source venv/bin/activate
+
+          # python-ruff doesn't work without rust
+          sed -i '/pytest-ruff/d' setup.cfg
+
+          pip install -e .[testing]
+      - name: Run tests
+        shell: msys2 {0}
+        run: |
+          source venv/bin/activate
+          pytest distutils/tests
+
   ci_setuptools:
     # Integration testing with setuptools
     if: ${{ false }}  # disabled for deprecation warnings
--- a/distutils/command/build_ext.py
+++ b/distutils/command/build_ext.py
@@ -21,7 +21,7 @@ from ..sysconfig import customize_compiler, get_python_version
 from ..sysconfig import get_config_h_filename
 from .._modified import newer_group
 from ..extension import Extension
-from ..util import get_platform
+from ..util import get_platform, is_mingw
 from distutils._log import log
 from . import py37compat
 
@@ -189,7 +189,7 @@ class build_ext(Command):
         # for extensions under windows use different directories
         # for Release and Debug builds.
         # also Python's library directory must be appended to library_dirs
-        if os.name == 'nt':
+        if os.name == 'nt' and not is_mingw():
             # the 'libs' directory is for binary installs - we assume that
             # must be the *native* platform.  But we don't really support
             # cross-compiling via a binary install anyway, so we let it go.
@@ -742,7 +742,7 @@ class build_ext(Command):
         # pyconfig.h that MSVC groks.  The other Windows compilers all seem
         # to need it mentioned explicitly, though, so that's what we do.
         # Append '_d' to the python import library on debug builds.
-        if sys.platform == "win32":
+        if sys.platform == "win32" and not is_mingw():
             from .._msvccompiler import MSVCCompiler
 
             if not isinstance(self.compiler, MSVCCompiler):
@@ -772,7 +772,7 @@ class build_ext(Command):
                 # A native build on an Android device or on Cygwin
                 if hasattr(sys, 'getandroidapilevel'):
                     link_libpython = True
-                elif sys.platform == 'cygwin':
+                elif sys.platform == 'cygwin' or is_mingw():
                     link_libpython = True
                 elif '_PYTHON_HOST_PLATFORM' in os.environ:
                     # We are cross-compiling for one of the relevant platforms
--- a/distutils/cygwinccompiler.py
+++ b/distutils/cygwinccompiler.py
@@ -57,7 +57,7 @@ def get_msvcr():
     try:
         msc_ver = int(match.group(1))
     except AttributeError:
-        return
+        return []
     try:
         return _msvcr_lookup[msc_ver]
     except KeyError:
@@ -277,7 +277,7 @@ class Mingw32CCompiler(CygwinCCompiler):
 
         self.set_executables(
             compiler='%s -O -Wall' % self.cc,
-            compiler_so='%s -mdll -O -Wall' % self.cc,
+            compiler_so='%s -shared -O -Wall' % self.cc,
             compiler_cxx='%s -O -Wall' % self.cxx,
             linker_exe='%s' % self.cc,
             linker_so='{} {}'.format(self.linker_dll, shared_option),
--- a/distutils/sysconfig.py
+++ b/distutils/sysconfig.py
@@ -18,6 +18,7 @@ import pathlib
 from .errors import DistutilsPlatformError
 from . import py39compat
 from ._functools import pass_none
+from .util import is_mingw
 
 IS_PYPY = '__pypy__' in sys.builtin_module_names
 
@@ -120,8 +121,10 @@ def get_python_inc(plat_specific=0, prefix=None):
     """
     default_prefix = BASE_EXEC_PREFIX if plat_specific else BASE_PREFIX
     resolved_prefix = prefix if prefix is not None else default_prefix
+    # MinGW imitates posix like layout, but os.name != posix
+    os_name = "posix" if is_mingw() else os.name
     try:
-        getter = globals()[f'_get_python_inc_{os.name}']
+        getter = globals()[f'_get_python_inc_{os_name}']
     except KeyError:
         raise DistutilsPlatformError(
             "I don't know where Python installs its C header files "
@@ -244,7 +247,7 @@ def get_python_lib(plat_specific=0, standard_lib=0, prefix=None):
         else:
             prefix = plat_specific and EXEC_PREFIX or PREFIX
 
-    if os.name == "posix":
+    if os.name == "posix" or is_mingw():
         if plat_specific or standard_lib:
             # Platform-specific modules (any module from a non-pure-Python
             # module distribution) or standard Python library modules.
@@ -273,7 +276,7 @@ def customize_compiler(compiler):  # noqa: C901
     Mainly needed on Unix, so we can plug in the information that
     varies across Unices and is stored in Python's Makefile.
     """
-    if compiler.compiler_type == "unix":
+    if compiler.compiler_type in ["unix", "cygwin", "mingw32"]:
         if sys.platform == "darwin":
             # Perform first-time customization of compiler-related
             # config vars on OS X now that we know we need a compiler.
--- a/distutils/tests/test_sysconfig.py
+++ b/distutils/tests/test_sysconfig.py
@@ -8,7 +8,7 @@ import pathlib
 import pytest
 import jaraco.envs
 import path
-from jaraco.text import trim
+from textwrap import dedent
 
 import distutils
 from distutils import sysconfig
@@ -17,6 +17,10 @@ from distutils.unixccompiler import UnixCCompiler
 from test.support import swap_item
 
 from . import py37compat
+
+
+def trim(s):
+    return dedent(s).strip()
 
 
 @pytest.mark.usefixtures('save_env')
--- a/distutils/tests/test_sysconfig.py
+++ b/distutils/tests/test_sysconfig.py
@@ -12,7 +12,7 @@ from jaraco.text import trim
 
 import distutils
 from distutils import sysconfig
-from distutils.ccompiler import get_default_compiler  # noqa: F401
+from distutils.ccompiler import new_compiler  # noqa: F401
 from distutils.unixccompiler import UnixCCompiler
 from test.support import swap_item
 
@@ -109,7 +109,7 @@ class TestSysconfig:
 
         return comp
 
-    @pytest.mark.skipif("get_default_compiler() != 'unix'")
+    @pytest.mark.skipif("not isinstance(new_compiler(), UnixCCompiler)")
     def test_customize_compiler(self):
         # Make sure that sysconfig._config_vars is initialized
         sysconfig.get_config_vars()
--- a/distutils/ccompiler.py
+++ b/distutils/ccompiler.py
@@ -19,7 +19,7 @@ from .spawn import spawn
 from .file_util import move_file
 from .dir_util import mkpath
 from ._modified import newer_group
-from .util import split_quoted, execute
+from .util import split_quoted, execute, is_mingw
 from ._log import log
 
 
@@ -1076,6 +1076,10 @@ def get_default_compiler(osname=None, platform=None):
         osname = os.name
     if platform is None:
         platform = sys.platform
+    # Mingw is a special case where sys.platform is 'win32' but we
+    # want to use the 'mingw32' compiler, so check it first
+    if is_mingw():
+        return 'mingw32'
     for pattern, compiler in _default_compilers:
         if (
             re.match(pattern, platform) is not None
--- a/distutils/tests/test_cygwinccompiler.py
+++ b/distutils/tests/test_cygwinccompiler.py
@@ -71,12 +71,12 @@ class TestCygwinCCompiler(support.TempdirManager):
         assert check_config_h()[0] == CONFIG_H_OK
 
     def test_get_msvcr(self):
-        # none
+        # []
         sys.version = (
             '2.6.1 (r261:67515, Dec  6 2008, 16:42:21) '
             '\n[GCC 4.0.1 (Apple Computer, Inc. build 5370)]'
         )
-        assert get_msvcr() is None
+        assert get_msvcr() == []
 
         # MSVC 7.0
         sys.version = (
--- a/distutils/tests/test_cygwinccompiler.py
+++ b/distutils/tests/test_cygwinccompiler.py
@@ -114,3 +114,10 @@ class TestCygwinCCompiler(support.TempdirManager):
         )
         with pytest.raises(ValueError):
             get_msvcr()
+
+    @pytest.mark.skipif('sys.platform != "cygwin"')
+    def test_dll_libraries_not_none(self):
+        from distutils.cygwinccompiler import CygwinCCompiler
+
+        compiler = CygwinCCompiler()
+        assert compiler.dll_libraries is not None
--- a/distutils/tests/test_install.py
+++ b/distutils/tests/test_install.py
@@ -16,6 +16,7 @@ from distutils.command.install import INSTALL_SCHEMES
 from distutils.core import Distribution
 from distutils.errors import DistutilsOptionError
 from distutils.extension import Extension
+from distutils.util import is_mingw
 
 from distutils.tests import support, missing_compiler_executable
 
@@ -120,7 +121,7 @@ class TestInstall(
         assert 'usersite' in cmd.config_vars
 
         actual_headers = os.path.relpath(cmd.install_headers, site.USER_BASE)
-        if os.name == 'nt':
+        if os.name == 'nt' and not is_mingw():
             site_path = os.path.relpath(os.path.dirname(orig_site), orig_base)
             include = os.path.join(site_path, 'Include')
         else:
--- a/distutils/tests/test_mingwccompiler.py
+++ b/distutils/tests/test_mingwccompiler.py
@@ -1,0 +1,45 @@
+import pytest
+
+from distutils.util import split_quoted, is_mingw
+from distutils.errors import DistutilsPlatformError, CCompilerError
+
+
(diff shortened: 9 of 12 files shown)


distutils/sysconfig.py
```diff
@@ -120,8 +120,11 @@ def get_python_inc(plat_specific=0, prefix=None):
     """
     default_prefix = BASE_EXEC_PREFIX if plat_specific else BASE_PREFIX
     resolved_prefix = prefix if prefix is not None else default_prefix
+    # MinGW imitates posix like layout, but os.name != posix
+    os_name = os.name if not sysconfig.get_platform().startswith("mingw") else "posix"
```
jaraco, line 124: Is there a reason this check doesn't use `is_mingw()`?

distutils/sysconfig.py
```diff
@@ -265,7 +268,7 @@ def customize_compiler(compiler):  # noqa: C901
     Mainly needed on Unix, so we can plug in the information that
     varies across Unices and is stored in Python's Makefile.
     """
-    if compiler.compiler_type == "unix":
+    if compiler.compiler_type in ["unix", "cygwin", "mingw32"]:
```
jaraco, line 271: I'd be okay with arbitrary changes to cygwinccompiler, but changes to sysconfig need a lot more scrutiny. This module is essentially deprecated and callers should prefer the stdlib `sysconfig` wherever possible. At the very least, it'd be great if there were some way to capture the expectation that's missed without these changes. Are there tests that could be run that fail on cygwin/mingw32 without these changes?

distutils/cygwinccompiler.py
```diff
@@ -57,6 +57,8 @@ def get_msvcr():
             return ['ucrt', 'vcruntime140']
         else:
             raise ValueError("Unknown MS Compiler version %s " % msc_ver)
+    else:
+        return []
```
jaraco, line 61: Here you change the return type of this method and the test that protects the return type, but it's unclear to me why without digging into it a bit. It would be nice if there was a test that would fail if self.dll_libraries is None.

distutils/sysconfig.py
```diff
@@ -120,8 +120,11 @@ def get_python_inc(plat_specific=0, prefix=None):
     """
     default_prefix = BASE_EXEC_PREFIX if plat_specific else BASE_PREFIX
     resolved_prefix = prefix if prefix is not None else default_prefix
+    # MinGW imitates posix like layout, but os.name != posix
+    os_name = os.name if not sysconfig.get_platform().startswith("mingw") else "posix"
+    inc_fun = f'_get_python_inc_{os_name}'
     try:
-        getter = globals()[f'_get_python_inc_{os.name}']
+        getter = globals()[inc_fun]
```
jaraco, line 127: I'm unsure why this changed; maybe bring back the inlined string?

distutils/sysconfig.py
jaraco, line 271: Maybe now that this value is selecting on multiple, it should be `isinstance(compiler, UnixCCompiler)`.

distutils/sysconfig.py
naveen521kk: This module didn't import `distutils.sysconfig` before and I thought that was done for a reason. If that's not the case, I'll use `is_mingw()` function.

distutils/sysconfig.py
naveen521kk: > At the very least, it'd be great if there were some way to capture the expectation that's missed without these changes.

When building extensions the compiler flags used while building python itself should be used, similar to how it's on UNIX platforms, and without these changes, those are missing.

> Are there tests that could be run that fail on cygwin/mingw32 without these changes?

Yeah, the compiler flags change and that could be tested. For example, with this change on a MINGW64 env I get the following output
```sh
running build_ext
building 'cwrapper' extension
creating build
creating build/temp.mingw_x86_64-cpython-310
gcc -Wno-unused-result -Wsign-compare -DNDEBUG -g -fwrapv -O3 -Wall -march=x86-64 -mtune=generic -O2 -pipe -O3 -march=x86-64 -mtune=generic -O2 -pipe -O3 -IC:/msys64/mingw64/include/python3.10 -c cwrapper.c -o build/temp.mingw_x86_64-cpython-310/cwrapper.o
writing build/temp.mingw_x86_64-cpython-310/cwrapper.cp310-mingw_x86_64.def
gcc -shared -Wl,--enable-auto-image-base -pipe -pipe -s build/temp.mingw_x86_64-cpython-310/cwrapper.o build/temp.mingw_x86_64-cpython-310/cwrapper.cp310-mingw_x86_64.def -LC:/msys64/mingw64/lib/python3.10/config-3.10 -LC:/msys64/mingw64/lib -lpython3.10 -o C:/dev/mingw-distutils/cext/cwrapper.cp310-mingw_x86_64.pyd
```
while without these changes I get
```sh
running build_ext
building 'cwrapper' extension
creating build
creating build/temp.mingw_x86_64-cpython-310
gcc -mdll -O -Wall -IC:/msys64/mingw64/include/python3.10 -c cwrapper.c -o build/temp.mingw_x86_64-cpython-310/cwrapper.o
writing build/temp.mingw_x86_64-cpython-310/cwrapper.cp310-mingw_x86_64.def
gcc -shared -s build/temp.mingw_x86_64-cpython-310/cwrapper.o build/temp.mingw_x86_64-cpython-310/cwrapper.cp310-mingw_x86_64.def -LC:/msys64/mingw64/lib/python3.10/config-3.10 -LC:/msys64/mingw64/lib -lpython3.10 -o C:/dev/mingw-distutils/cext/cwrapper.cp310-mingw_x86_64.pyd
```

> Maybe now that this value is selecting on multiple, it should be isinstance(compiler, UnixCCompiler).

👍

distutils/sysconfig.py
jaraco: I don't see any protections against importing it (e.g. a comment indicating not to import other modules), so I'd say it's safe. Let's re-use it. Maybe something like this:

```diff
diff --git a/distutils/sysconfig.py b/distutils/sysconfig.py
index 35e5289..0fbb2e1 100644
--- a/distutils/sysconfig.py
+++ b/distutils/sysconfig.py
@@ -18,6 +18,7 @@ import pathlib
 from .errors import DistutilsPlatformError
 from . import py39compat
 from ._functools import pass_none
+from .util import is_mingw
 
 IS_PYPY = '__pypy__' in sys.builtin_module_names
 
@@ -121,7 +122,7 @@ def get_python_inc(plat_specific=0, prefix=None):
     default_prefix = BASE_EXEC_PREFIX if plat_specific else BASE_PREFIX
     resolved_prefix = prefix if prefix is not None else default_prefix
     # MinGW imitates posix like layout, but os.name != posix
-    os_name = os.name if not sysconfig.get_platform().startswith("mingw") else "posix"
+    os_name = "posix" if is_mingw() else os.name
     try:
         getter = globals()[f'_get_python_inc_{os_name}']
     except KeyError:
```

distutils/sysconfig.py
jaraco: I meant to ask for tests that could be added to the test suite that capture this (missed/fixed) expectation.

jaraco:
I notice there are conflicts. I tried to resolve the conflicts, but I don't seem to have access to the fork, so I've pushed a couple of commits to the [add-mingw-support-jaraco branch](/pypa/distutils/tree/add-mingw-support-jaraco), resolving the conflict. Feel free to pull those changes in.

naveen521kk:
Thanks, I would be able to work on this later this week.

distutils/sysconfig.py
naveen521kk: `compiler.compiler_type` will never be `cygwin`, see https://github.com/pypa/distutils/issues/185, I'll better do a instance check instance.

naveen521kk: https://github.com/pypa/distutils/pull/184/commits/800fae90cb8171304c0a59f2157ebd52b005845d. There already exist tests to test the behaviour of this function and I've made it run on MinGW too.

naveen521kk: Changing to `isinstance` check fails the test suite. The `compiler` passed from the test suite is *not* a subclass of `UnixCCompiler` and fails this condition.

Karimsultan:
> See individual commits for details on what this PR changes.
> 
> 
> 
> A new version of https://github.com/pypa/distutils/pull/78
> 
> Fixes https://github.com/pypa/distutils/issues/34
> 
> 
> 
> cc @lazka

Karimsultan (approved):

cr1901:
@naveen521kk What is the status of this PR? I've been switching to PDM for a lot of my work, and installing a number of packages break with the good ol' `--plat-name must be one of ('win32', 'win-amd64', 'win-arm32', 'win-arm64')` error.

Since the [workaround](https://github.com/pypa/setuptools/issues/3625) is going away, soon this means I won't be able to install packages anymore.

naveen521kk:
> @naveen521kk What is the status of this PR? I've been switching to PDM for a lot of my work, and installing a number of packages break with the good ol' --plat-name must be one of ('win32', 'win-amd64', 'win-arm32', 'win-arm64') error.

Hey, I'm sorry, I kinda forgot about this PR. I'll spend some time this week week to get this working.

cr1901:
@naveen521kk

>Hey, I'm sorry, I kinda forgot about this PR. 

No worries, I'm kinda happy you responded at all, tbh.

>I'll spend some time this week to get this working.

Excellent, tyvm. I'm happy to test locally, although I'm not sure how I would replace the vendored version in my Python with this version for testing purposes. Right now I'm still `pip install`ing `pdm` from the source tree b/c of the `--plat-name` errors (it works otherwise fine). Will be interested in whether this allows installing `pdm` from PyPI to succeed.

.github/workflows/main.yml
```diff
@@ -120,6 +120,46 @@ jobs:
...
+    strategy:
+      matrix:
+        include:
+          - { sys: mingw64, env: x86_64 }
+          - { sys: mingw32, env: i686 }
+          - { sys: ucrt64,  env: ucrt-x86_64 }
+          - { sys: clang64, env: clang-x86_64 }
+    runs-on: windows-latest
+    steps:
+      - uses: actions/checkout@v2
```
lazka, line 133:
```suggestion
      - uses: actions/checkout@v4
```

distutils/command/build_ext.py
```diff
@@ -217,14 +217,18 @@ def finalize_options(self):  # noqa: C901
                 new_lib = os.path.join(new_lib, suffix)
             self.library_dirs.append(new_lib)
 
-        # For extensions under Cygwin, Python's library directory must be
+        # For extensions under Cygwin and MinGW, Python's library directory must be
         # appended to library_dirs
-        if sys.platform[:6] == 'cygwin':
+        if sys.platform[:6] == 'cygwin' or is_mingw():
             if not sysconfig.python_build:
+                config_dir_name = os.path.basename(sysconfig.get_config_var('LIBPL'))
```
lazka, line 224: Is this still needed with the recent relocation fix i.e. could we use LIBPL as is?

distutils/sysconfig.py
```diff
@@ -244,7 +247,7 @@ def get_python_lib(plat_specific=0, standard_lib=0, prefix=None):
         else:
             prefix = plat_specific and EXEC_PREFIX or PREFIX
 
-    if os.name == "posix":
+    if os.name == "posix" or sysconfig.get_platform().startswith('mingw'):
```
lazka, line 250: is is_mingw() not used on purpose here?

distutils/tests/test_sysconfig.py
```diff
@@ -8,17 +8,21 @@
 import pytest
 import jaraco.envs
 import path
-from jaraco.text import trim
```
lazka, line 11: Just wondering, which dependency was the problem? (I guess I could try)

Either way it seems like a good trade to lose a dependency, but ymmv of course.

lazka:
Should we also include https://github.com/msys2-contrib/cpython-mingw/pull/157 ?

distutils/tests/test_sysconfig.py
lazka: ah, pydantic-core gets pulled in. It builds fine here if I install rust, but it takes quite a bit to build (some minutes)

distutils/command/build_ext.py
naveen521kk: Yeah, it'll work for mingw, but I don't know whether it would work for Cygwin.

distutils/sysconfig.py
naveen521kk: not really, I'll change it.

distutils/command/build_ext.py
lazka: On cygwin it returns `'/usr/lib/python3.9/config-3.9-x86_64-cygwin'` and `'/usr/lib/python3.9/config'` doesn't exist. tbh I don't quite understand what this is doing in the first place..

naveen521kk: I'm also not sure, I copied it from the old patches. Maybe we don't need it at all?

naveen521kk:
> Should we also include https://github.com/msys2-contrib/cpython-mingw/pull/157 ?

I think so, I'll include them.

distutils/command/build_ext.py
lazka: Without having tested it, only the build_python case makes somewhat sense, but since this version of distutils isn't used for building CPython internal extensions https://github.com/pypa/setuptools/commit/336facde9301e041b2e33a7b46fe1d37991a7ed3 I think it might not be needed at all, yes.

naveen521kk: > Without having tested it, only the build_python case makes somewhat sense, but since this version of distutils isn't used for building CPython internal extensions https://github.com/pypa/setuptools/commit/336facde9301e041b2e33a7b46fe1d37991a7ed3 I think it might not be needed at all, yes.

I removed this part of the patch and things seems to be working, so I guess it's all fine.

naveen521kk:
This PR is ready for review!

naveen521kk:
Hey, @jaraco it's been quite long, can you review this PR?

jaraco:
I've resolved the conflicts and I'd like to test the changes, but GitHub is rejecting me pushing the resolution.

```
 distutils add-mingw-support @ git push
fatal: The upstream branch of your current branch does not match
the name of your current branch.  To push to the upstream branch
on the remote, use

    git push origin HEAD:refs/pull/184/head

To push to the branch of the same name on the remote, use

    git push origin HEAD

To avoid automatically configuring an upstream branch when its name
won't match the local branch, see option 'simple' of branch.autoSetupMerge
in 'git help config'.

 distutils add-mingw-support [128] @ git push origin HEAD:refs/pull/184/head
Enumerating objects: 99, done.
Counting objects: 100% (99/99), done.
Delta compression using up to 12 threads
Compressing objects: 100% (66/66), done.
Writing objects: 100% (68/68), 10.07 KiB | 10.07 MiB/s, done.
Total 68 (delta 51), reused 0 (delta 0), pack-reused 0 (from 0)
remote: Resolving deltas: 100% (51/51), completed with 25 local objects.
To https://github.com/pypa/distutils
 ! [remote rejected]     HEAD -> refs/pull/184/head (deny updating a hidden ref)
error: failed to push some refs to 'https://github.com/pypa/distutils'
```

I'm not sure what "deny updating a hidden ref" means. I've pushed the commit (1b38bc903) to the add-mingw-support branch to see if tests pass.

jaraco:
That failed because there was a latent reference to setup.cfg. Corrected in 416ef6a75.

jaraco:
And another revision to replace `testing` with `test`. 5fb38a1e1.

jaraco:
Tests are now failing during collection:

```
______________ ERROR collecting distutils/tests/test_install.py _______________
ImportError while importing test module 'D:/a/distutils/distutils/distutils/tests/test_install.py'.
Hint: make sure your test modules/packages have valid Python names.
Traceback:
distutils/tests/test_install.py:16: in <module>
    from distutils.util import is_mingw
E   ImportError: cannot import name 'is_mingw' from 'distutils.util' (D:/a/_temp/msys64/mingw64/lib/python3.11/distutils/util.py)
___________ ERROR collecting distutils/tests/test_mingwccompiler.py ___________
ImportError while importing test module 'D:/a/distutils/distutils/distutils/tests/test_mingwccompiler.py'.
Hint: make sure your test modules/packages have valid Python names.
Traceback:
distutils/tests/test_mingwccompiler.py:3: in <module>
    from distutils.util import split_quoted, is_mingw
E   ImportError: cannot import name 'is_mingw' from 'distutils.util' (D:/a/_temp/msys64/mingw64/lib/python3.11/distutils/util.py)
___________ ERROR collecting distutils/tests/test_unixccompiler.py ____________
ImportError while importing test module 'D:/a/distutils/distutils/distutils/tests/test_unixccompiler.py'.
Hint: make sure your test modules/packages have valid Python names.
Traceback:
distutils/tests/test_unixccompiler.py:10: in <module>
    from distutils.util import _clear_cached_macosx_ver
E   ImportError: cannot import name '_clear_cached_macosx_ver' from 'distutils.util' (D:/a/_temp/msys64/mingw64/lib/python3.11/distutils/util.py)
```

Clearly the wrong 'distutils' is getting imported. I don't know why.

jaraco:
@naveen521kk Can you pull 5fb38a1e1 over to your branch so we can get the CI to update here and then investigate why the wrong distutils is getting imported (but only for the mingw jobs)?

abravalheri:
> Clearly the wrong 'distutils' is getting imported. I don't know why.

What is the order in `sys.meta_path`? Anything weird there?

lazka:
> @naveen521kk Can you pull [5fb38a1](https://github.com/pypa/distutils/commit/5fb38a1e169dd17103b1abe9282d3dae5f049a9a) over to your branch so we can get the CI to update here and then investigate why the wrong distutils is getting imported (but only for the mingw jobs)?

From what I see you succeeded in pushing it before writing this (??). I can also rebase the whole thing on top of main if wanted.

lazka:
> Clearly the wrong 'distutils' is getting imported. I don't know why.

I can reproduce locally. I'll try to figure out why.

lazka:
> What is the order in `sys.meta_path`? Anything weird there?

```python
distutils/tests/test_install.py:8: in <module>
    assert 0, sys.meta_path
E   AssertionError: [<_pytest.assertion.rewrite.AssertionRewritingHook object at 0x000001a442ffe510>, <_virtualenv._Finder object at 0x000...mportlib_external.PathFinder'>, <class '__editable___distutils_0_1_dev4112_g5fb38a1_d20240627_finder._EditableFinder'>]
```

lazka:
Removing this line makes things work again: https://github.com/pypa/distutils/blob/4fd2d569e37268b8c23f3cc45879465f9c269d92/pytest.ini#L5

naveen521kk:
> @naveen521kk Can you pull [5fb38a1](https://github.com/pypa/distutils/commit/5fb38a1e169dd17103b1abe9282d3dae5f049a9a) over to your branch so we can get the CI to update here and then investigate why the wrong distutils is getting imported (but only for the mingw jobs)?

I've pulled those changes. I'll also check out why the wrong distutils are imported.

> What is the order in sys.meta_path? Anything weird there?

Not sure if there's anything weird there but this is what I get.
```
$ python
Python 3.11.9 (main, Apr 12 2024, 09:55:31)  [GCC 13.2.0 64 bit (AMD64)] on win32
Type "help", "copyright", "credits" or "license" for more information.
>>> import sys
>>> sys.meta_path
[<_virtualenv._Finder object at 0x000001745b6ffe50>, <class '_frozen_importlib.BuiltinImporter'>, <class '_frozen_importlib.FrozenImporter'>, <class '_frozen_importlib_external.PathFinder'>, <class '__editable___distutils_0_1_dev4112_g5fb38a1_d20240627_finder._EditableFinder'>]
>>>

```

jaraco:
I encountered similar errors in collection in #259, which I worked around by pinning pytest. And I can see the pin has taken effect, as pytest 8.0.2 is being used. Yet, somehow the import mode is still finding the wrong distutils. I did a deep dive on the issue in https://github.com/pytest-dev/pytest/issues/12490.  I still have no clue how the msys environment is somehow reviving that issue.

lazka:
How is that supposed to work anyway? With the pytest setting the current dir isn't in sys.path anymore. The stdlib pathfinder is before the editable finder in sys.meta_path, so I kinda expect it to pick up the wrong distutils, no?

I need to check in a Linux with Python 3.11 somehow, I only have 3.12 without distutils around.

jaraco:
When I was testing with pytest 8.2, where `_pytest.pathlib.resolve_pkg_root_and_module_name`, invoked when `import-mode=importlib`, doesn't rely on sys.path at all. On Pytest 8.0.2, the logic is very different. Let me see if I can trace why and how the import works on my macOS machine.

lazka:
I now tried on Ubuntu 22.04, using Python 3.10, and it fails the same way, importing the wrong distutils.

```
python3 -m venv venv
source venv/bin/activate
pip install -e '.[test]'
pytest distutils/tests
```

jaraco:
Aha. I see the difference now. It's the invocation of `pytest distutils/tests`.

I created this dockerfile to replicate the behavior.

```Dockerfile
from jaraco/multipy-tox:jammy

RUN git clone https://github.com/msys2-contrib/distutils
WORKDIR distutils
RUN git checkout add-mingw-support
RUN py -3.10 -m venv .venv
RUN py -m pip install -e .[test]
# simulate activating the venv:
ENV PATH=.venv/bin:$PATH
CMD pytest distutils/tests
```

If I run that docker image, it fails with the same issues in CI. If I instead run docker with the command `bash -c pytest`, the collection completes and the tests run (with a few failures not yet seen).

I think here's what's happening: 

When passing `distutils/tests` to pytest, it causes `distutils/*` not to be collected, and thus falls back to sys.path for resolving the `import distutils`. If not passing `distutils/tests`, pytest will traverse `./distutils` and because it's also run with `--doctest-modules`, the modules in `distutils/*` get imported during collection.

Indeed, if I disable the line with `--doctest-modules` and then run tests with `pytest`, the same collection failure occurs. The same problem exists even after restoring pytest 8.2+.

It seems that the reason I haven't encountered this issue previously or with other skeleton-based projects is because I rarely ever direct the collection by providing a subpath, but instead I use `-k <pattern>` to downselect tests after collection (so doctests are always collected even if not run).

> What about [adding some config]?

I really try to avoid having to add config, especially to a single project. I'd like to be able to rely on defaults and best practices to make the project under test importable. I guess distutils is special here, because although it's pip-installed, a pip-installed package doesn't take precedence over a stdlib module.

How about for now, let's update the test execution to run the same way as it runs for other jobs (without specifying `distutils/tests`)?

lazka:
> How about for now, let's update the test execution to run the same way as it runs for other jobs (without specifying `distutils/tests`)?

Done, see the commits for details.

It seems a bit brittle since this import side effect no longer seems to be there with pytest 8.2.0 for example, but fine with me. If you update pytest in the future I'd be happy to help debug Windows specific issues.

For future reference, this makes the tests pass with pytest 8.2.0 here:

```diff
diff --git a/pytest.ini b/pytest.ini
index b53e0d93..14188d07 100644
--- a/pytest.ini
+++ b/pytest.ini
@@ -3,7 +3,7 @@ norecursedirs=dist build .tox .eggs
 addopts=
 	--doctest-modules
 	--import-mode importlib
-consider_namespace_packages=true
+pythonpath= .
 filterwarnings=
 	## upstream
 ```

jaraco:
Looks great. Thanks!

lazka:
Thanks!

If there is any problems/issues feel free to ping us

PR #191, opened 2022-11-20
Apply refurb suggestions

DimitriPapadopoulos:
* [FURB123]: Replace `str(x)` with `x`
* [FURB113]: Use `x.extend(...)` instead of repeatedly calling `x.append()`
* [FURB108]: Use `x in (y, z)` instead of `x == y or x == z`
* [FURB131]: Replace `del x[y]` with `x.pop(y)`
* [FURB109]: Replace `in [x, y, z]` with `in (x, y, z)`

### DIFF
--- a/distutils/_collections.py
+++ b/distutils/_collections.py
@@ -185,7 +185,7 @@ class RangeMap(dict):
         return (sorted_keys[RangeMap.first_item], sorted_keys[RangeMap.last_item])
 
     # some special values for the RangeMap
-    undefined_value = type(str('RangeValueUndefined'), (), {})()
+    undefined_value = type('RangeValueUndefined', (), {})()
 
     class Item(int):
         "RangeMap Item"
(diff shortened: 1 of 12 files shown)


distutils/bcppcompiler.py
```diff
@@ -294,8 +294,7 @@ def link(  # noqa: C901
                     ld_args.append(libfile)
 
             # some default libraries
-            ld_args.append('import32')
-            ld_args.append('cw32mt')
+            ld_args.extend(('import32', 'cw32mt'))
```
jaraco, line 297: I slightly prefer the readability of the other syntax, but this form is okay, and apparently preferred by refurb.

distutils/_collections.py
```diff
@@ -185,7 +185,7 @@ def bounds(self):
         return (sorted_keys[RangeMap.first_item], sorted_keys[RangeMap.last_item])
 
     # some special values for the RangeMap
-    undefined_value = type(str('RangeValueUndefined'), (), {})()
```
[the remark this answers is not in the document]
jaraco: Applied as jaraco/jaraco.collections@5a40782.

distutils/bcppcompiler.py
DimitriPapadopoulos: Agreed, I'm not always happy with refurb suggestions.

PR #192, opened 2022-11-22
Backfill `distutils.log.Log` for compatibility

abravalheri:
in pypa/setuptools#3693 we learned that some projects rely on `distutils.log.Log` and users are asking for `distutils.log.Log` to be added back to ensure compatibility is not broken.

I am not sure what is the best way forward, but I am providing this PR if pypa/distutils' maintainers are interested in the "backfill" approach.

* I found just a few usages in https://grep.app/search?q=from%20distutils.log%20import%20Log,
  * However it seams that the main problem is users importing `numpy.disutils` without pinning setuptools.
* I am not sure about adding tests. I noticed that all the tests about deprecated `distutils.log` behaviour were removed.

### DIFF
--- a/distutils/log.py
+++ b/distutils/log.py
@@ -40,7 +40,7 @@ def set_verbosity(v):
 
 
 class Log(logging.Logger):
-    """distutils.log.Logger is deprecated, please use an alternative from `logging`."""
+    """distutils.log.Log is deprecated, please use an alternative from `logging`."""
 
     def __init__(self, threshold=WARN):
         warnings.warn(Log.__doc__)  # avoid DeprecationWarning to ensure warn is shown
--- a/distutils/log.py
+++ b/distutils/log.py
@@ -5,6 +5,7 @@ Retained for compatibility and should not be used.
 """
 
 import logging
+import warnings
 
 from ._log import log as _global_log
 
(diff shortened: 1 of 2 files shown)


abravalheri:
The error in `ci_setuptools` is probably solved by https://github.com/pypa/setuptools/pull/3701

rgommers:
Thanks @abravalheri, this PR is much appreciated.

> However it seams that the main problem is users importing `numpy.disutils` without pinning setuptools.

This is also for usage at runtime, and we've got issues in e.g. Debian which simply ships only one `setuptools` version, so there is nothing to pin. Released `numpy` versions are out there using `Log`, can't yank those. So this seems like a good way to keep things from falling apart. Can be done without tests imho, it's just a band-aid.

abravalheri:
Thank you very much @rgommers.

The way I manually tested this was:

```bash
git clone https://github.com/abravalheri/distutils --branch setuptools-issue-3693
git clone https://github.com/pypa/setuptools
rm -rf setuptools/setuptools/_distutils
cp -r distutils/distutils setuptools/setuptools/_distutils
virtualenv .venv
.venv/bin/python -m pip install -U pip ./setuptools
.venv/bin/python -m pip install -m numpy
.venv/bin/python -c '
import distutils

assert "setuptools/_distutils" in next(iter(distutils.__path__))

from numpy.distutils import log

log.warn("hello world")

import logging
assert isinstance(log.Log(), logging.Logger), f"{log.Log.__mro__=}"
'
```

<details>
<summary>After that I can see a coloured `hello world` message, so I am assuming the patch works well.</summary>

image
</details>

jaraco:
Brilliant. Thank you.

PR #200, opened 2023-02-06
For 'get_python_inc', bypass missing config include dir

jaraco:
- Add xfail test capturing missed expectation. Ref pypa/distutils#178.
- In _get_python_inc_posix, only return an extant path from the sysconfig. Fixes pypa/distutils#178.

### DIFF
--- a/distutils/_msvccompiler.py
+++ b/distutils/_msvccompiler.py
@@ -339,7 +339,6 @@ class MSVCCompiler(CCompiler):
         extra_postargs=None,
         depends=None,
     ):
-
         if not self.initialized:
             self.initialize()
         compile_info = self._setup_compile(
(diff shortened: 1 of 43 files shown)


mattip:
What would your feeling be about adding pypy to the [CI matrix](https://github.com/pypa/distutils/blob/b504971d65113b0fa603bf55199e2e032f041f8e/.github/workflows/main.yml#L18)? On the one hand it would be good to test everything, on the other PyPy is rather niche, things may break, and it could slow down development. The [python action](https://github.com/actions/setup-python#basic-usage) accepts PyPy versions like `'pypy3.9'`

jaraco:
I'm not opposed, and it would probably be worthwhile to have at least one PyPy in the matrix.

PR #201, opened 2023-02-06
add a pypy CI run

mattip:
add a PyPy3.9 run on ubuntu


mattip:
"luckily" the PyPy run will probably be quite slow, so any spurious failures should not interfere and cancel the CI runs of other matrix configurations. PyPy's native interpreter is about 2x slower than CPython, and typically CI testing is a worst-case scenario for the JIT.

Issue #185, opened 2022-10-29
`CygwinCCompiler` is never used


naveen521kk:
While working on https://github.com/pypa/distutils/pull/184/files#r1008651830 I found that `CygwinCCompiler` is never used. If it was used it should be failing with `TypeError` as there's something wrong with `get_msvcr` (it should be returning `None`). 

It uses `UnixCCompiler` instead while compiling on that platform.
```sh
$ python -c 'import distutils.ccompiler as a; print(a.new_compiler())'
<distutils.unixccompiler.UnixCCompiler object at 0x6ffffff07c10>
```

I think the below line should be changed to `cygwin` instead of `unix` to use the correct compiler classes. Not sure why it was set to `unix` though. (and the blame is 22 years old, so no idea https://github.com/pypa/distutils/commit/e49a11577577efa7493fe079aec5b257a1796d00)

https://github.com/pypa/distutils/blob/6c39b5053c17ec41690dab9d11399e2f39389892/distutils/ccompiler.py#L1024

radarhere:
You may be interested to know that https://github.com/pypa/distutils/pull/209#issuecomment-1554669584 suggested dropping CygwinCCompiler instead.

jaraco:
What's the status of this issue now that #209 is merged?

jaraco:
Should we delete the CygwinCCompiler?

PR #209, opened 2023-05-14
Port code from CygwinCCompiler to UnixCCompiler

DWesl:
https://github.com/python-pillow/Pillow/issues/7158#issuecomment-1546746716 and 
https://github.com/pypa/setuptools/blob/e76db4408e40a5c0a218ef6f652454f69c5aca59/setuptools/_distutils/ccompiler.py#L1055-L1058 
suggest that Cygwin uses `UnixCCompiler` rather than `CygwinCCompiler` by default, so `UnixCCompiler` would also need to know how to find shared libraries, import libraries, and static libraries on Cygwin.

Port of #139 to `UnixCCompiler`


DWesl:
`UnixCCompiler.find_library_file` now returns library names that are likely to occur on Cygwin, rather than those likely on Linux, which causes the doctest to fail.  Should I fix that with `# doctest: ELLIPSIS` or something like
```python
>>> result = compiler.find_library_file(dirs, 'abc').replace('\\', '/')
>>> if sys.platform == "cygwin":
...     print(result == "/foo/bar/existing/cygabc.dll")
... else:
...     print(result == "/foo/bar/existing/libabc.dylib")
True
```

DWesl:
Given [Cygwin seems to use `UnixCCompiler` rather than `CygwinCCompiler`](https://github.com/pypa/setuptools/blob/e76db4408e40a5c0a218ef6f652454f69c5aca59/setuptools/_distutils/ccompiler.py#L1056-L1058), should we also drop `cygwinccompiler.py`?

radarhere:
> Given [Cygwin seems to use `UnixCCompiler` rather than `CygwinCCompiler`](https://github.com/pypa/setuptools/blob/e76db4408e40a5c0a218ef6f652454f69c5aca59/setuptools/_distutils/ccompiler.py#L1056-L1058), should we also drop `cygwinccompiler.py`?

#185 similarly feels CygwinCCompiler isn't used, although thinks it should be.

DWesl:
The odd doctests suggest having `CygwinCCompiler` be a thin wrapper around `UnixCCompiler` might be the way to go: the only thing I can think to change would be the doctests in this PR, as everything else seems fine.

distutils/unixccompiler.py
```diff
@@ -141,6 +141,10 @@ class UnixCCompiler(CCompiler):
     xcode_stub_lib_format = dylib_lib_format
     if sys.platform == "cygwin":
         exe_extension = ".exe"
+        shared_lib_extension = ".dll.a"
+        dylib_lib_extension = ".dll"
+        static_lib_format = shared_lib_format = "lib%s%s"
```
DWesl, line 146:
```suggestion
```
The pattern here is the same as on Linux, just with different extensions.  We can remove this and rely on the definition on line 140.

jaraco:
Oh, heck. Tests on cygwin are currently disabled because ruff fails to install there.

DWesl:
There's some progress on rust-lang/rust#79854; I'm somewhat curious how well the work done to cross-compile from Linux would carry over to cross-compiling from Windows.

ookiineko:
>  I'm somewhat curious how well the work done to cross-compile from Linux would carry over to cross-compiling from Windows.

there wasn't any existing cygwin cross compiler available on mingw or msys, but there is [one on fedora linux](https://copr.fedorainfracloud.org/coprs/yselkowitz/cygwin/)

i also tried to get a cygwin cross compiler on archlinux: [arch-cygwin](https://github.com/ookiineko/arch-cygwin) (unofficial), it uses PKGBUILD (which is the same as what MSYS2 is using), might be useful if someone want to build them and package on msys2

jaraco:
> How should I be fixing the doctests to not fail with working code on Cygwin?

I don't see any problem with the doctests. Perhaps they've been resolved.

The only failing test now is the failing coverage test, which is spurious and only there because the coverage is only checked on one platform, so it's ignorable.

DWesl:
> > How should I be fixing the doctests to not fail with working code on Cygwin?
> 
> I don't see any problem with the doctests. Perhaps they've been resolved.

It was a failure on the Cygwin doctests, which I think got disabled because there's no Rust compiler for Cygwin

Kreijstal:
> > > How should I be fixing the doctests to not fail with working code on Cygwin?
> > 
> > 
> > I don't see any problem with the doctests. Perhaps they've been resolved.
> 
> It was a failure on the Cygwin doctests, which I think got disabled because there's no Rust compiler for Cygwin

the problem is usually path conversion
```
https://github.com/thepowersgang/mrustc
```
you can kinda get rust here on cygwin, but there are many problems with path conversion and cygwin not being linux

PR #211, opened 2023-08-06
CI: Install git on Cygwin CI runner

DWesl:
Cygwin pip now has a chance to resolve everything on the command line.

Git commands will fail because the repository was created by Windows git, but this should work until the above problems are solved.

### DIFF
--- a/.coveragerc
+++ b/.coveragerc
@@ -1,0 +1,9 @@
+[run]
+omit =
+	# leading `*/` for pytest-dev/pytest-cov#456
+	*/.tox/*
+disable_warnings =
+	couldnt-parse
+
+[report]
+show_missing = True
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,19 @@
+root = true
+
+[*]
+charset = utf-8
+indent_style = tab
+indent_size = 4
+insert_final_newline = true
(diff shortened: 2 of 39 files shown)


DWesl:
It's possible to fix the actions/checkout post-run failure, but that doesn't seem to be affecting anything, so I'm leaving it unless someone tells me otherwise.

DWesl:
I don't understand why 3.7 fails to install the pytest branch, but I suspect that's unrelated.

DWesl:
It seems pytest dropped support for Python 3.7 before the fix for the MRO bug went in:
https://github.com/RonnyPfannschmidt/pytest/blob/fix-10447-maker-mro-order-needs-reverse/setup.cfg#L53
Should that batch of tests get bumped to using Python 3.8 instead?

DWesl:
The banner at the top of https://docs.python.org/3.7/ suggests that 3.7 is EOL, which also suggests moving the failing CI runs to 3.8.

jaraco:
Unfortunately, cygwin got disabled due to incompatibility with Rust dependencies. I was hopeful that the pinning of inflect might help, so I've updated it per the suggestion above, but it looks like the additional reliance on `ruff` is probably preventing re-enabling Cygwin.

PR #213, opened 2023-08-30
"Fix" interoperability of `rfc822_escape` with stblib's `email` package

abravalheri:
In pypa/setuptools#4033 we learned that there are some "interoperability challenges" regarding the way distutil's implementation of RFC822 escaping works and the way stdlib's `email` package works (which is in turn used by `pypa/wheel` and probably other `pypa` projects).

While there is no bug in the existing implementation of `pypa/distutils` *per se*[^1], when someone attempts to re-generate the `METADATA` file using the stdlib (as in `pypa/wheel`), the resulting file can make the parser choke on control characters that are interpreted by Python as line breaks.

I am proposing this PR in `pypa/distutils` because I think it would be nice to make the `rfc822_escape` function more robust.
However, if the maintainers of `pypa/distutils` believe that this is not the best place for tackling the issue (since the produced files are not technically broken), there are 2 alternatives: (a) re-implement `rfc822_escape` in `setuptools` and patch if necessary, (b) discuss with `pypa/wheel` if the problem can be handled there[^2]. Please let me know your suggestion.

### Implemented changes
- [Modified `test_utils.test_rfc822_escape`, capturing interoperability requirements](https://github.com/pypa/distutils/commit/222b249f4f7ee9c1b2fae7f483db88c031fe4302)
- [Added `test_dist.TestMetadata.test_round_trip_through_email_generator`, capturing interoperability requirements](https://github.com/pypa/distutils/commit/157fbfed51a405866c9f63cc75c69cfac6b8735e)
- ["Fix"/improve interoperability of `distutils.util.rfc822_escape` with stblib's `email` library](https://github.com/pypa/distutils/commit/0ece9871247625ed3541b66529ca654039a5d8b5)

[^1]: The generated `PKG-INFO` files parse just fine...
[^2]: The file that actually fails to parse correctly is produced in [`bdist_wheel`](https://github.com/pypa/wheel/blob/0.41.2/src/wheel/bdist_wheel.py#L587)

### DIFF
--- a/distutils/tests/test_dist.py
+++ b/distutils/tests/test_dist.py
@@ -1,6 +1,9 @@
 """Tests for distutils.dist."""
 import os
 import io
+import email
+import email.policy
+import email.generator
 import sys
 import warnings
 import textwrap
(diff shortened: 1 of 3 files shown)


abravalheri:
There seems to be something unrelated happening on the failing tests:

```console:
test_cygwin - tox.py:install_deps
   ERROR: Error [Errno 2] No such file or directory: 'git' while executing command git version
test (~3.7.0-0, ubuntu-latest - tox.py:install_deps (timeout while installing test dependencies?)
   py: exit 2 (737.65 seconds) /home/runner/work/distutils/distutils> python -I -m pip install docutils 'flake8<5' 'jaraco.envs>=2.4' jaraco.path jaraco.text more_itertools path pyfakefs pytest-black pytest-cov 'pytest-enabler>=1.3' pytest-flake8 'pytest@ git+https://github.com/RonnyPfannschmidt/pytest@fix-10447-maker-mro-order-needs-reverse' pid=1821
```

jaraco (approved):

PR #214, opened 2023-09-26
GNU: use -Wl,-rpath,<dir> instead of -Wl,-R<dir>

haampie:
The latter is supported in binutils for backwards compatibility, but in
general `-R<path>` is equivalent to `--just-symbols=<path>` when `path`
is a file; only when it's a directory, it's treated as `-rpath=<path>`.

Better avoid that ambiguity and use `-rpath`.

Also split `-Wl,--enable-new-dtags` and `-Wl,-rpath,...` into two
separate arguments, which is more common, and more likely to be parsed
correctly by compiler wrappers.

This commit does not attempt to add `--enable-new-dtags` to other
linkers than binutils ld/gold that support the flag.

### DIFF
--- a/distutils/tests/test_unixccompiler.py
+++ b/distutils/tests/test_unixccompiler.py
@@ -189,7 +189,7 @@ class TestUnixCCompiler(support.TempdirManager):
         sysconfig.get_config_var = gcv
         assert self.cc.rpath_foo() == [
             '-Wl,--enable-new-dtags',
-            '-Wl,-rpath,/foo'
+            '-Wl,-rpath,/foo',
         ]
 
         # non-GCC GNULD
--- a/distutils/tests/test_unixccompiler.py
+++ b/distutils/tests/test_unixccompiler.py
@@ -186,7 +186,10 @@ class TestUnixCCompiler(support.TempdirManager):
                 return 'yes'
 
         sysconfig.get_config_var = gcv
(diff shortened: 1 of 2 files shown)


jaraco:
After fixing the linter failures, there's still a test failure:

```
>       assert self.cc.rpath_foo() == '-Wl,--enable-new-dtags,-R/foo'
E       AssertionError: assert ['-Wl,--enable-new-dtags', '-Wl,-rpath,/foo'] == '-Wl,--enable-new-dtags,-R/foo'
E        +  where ['-Wl,--enable-new-dtags', '-Wl,-rpath,/foo'] = <bound method compiler_wrapper.<locals>.CompilerWrapper.rpath_foo of <distutils.tests.test_unixccompiler.compiler_wrapper.<locals>.CompilerWrapper object at 0x11a03b920>>()
E        +    where <bound method compiler_wrapper.<locals>.CompilerWrapper.rpath_foo of <distutils.tests.test_unixccompiler.compiler_wrapper.<locals>.CompilerWrapper object at 0x11a03b920>> = <distutils.tests.test_unixccompiler.compiler_wrapper.<locals>.CompilerWrapper object at 0x11a03b920>.rpath_foo
E        +      where <distutils.tests.test_unixccompiler.compiler_wrapper.<locals>.CompilerWrapper object at 0x11a03b920> = <distutils.tests.test_unixccompiler.TestUnixCCompiler object at 0x118158c80>.cc
```

PR #220, opened 2023-12-12
#219 Use sysconfig.get_config_h_filename() to locate pyconfig.h

### LINKED ISSUE #219: _get_python_inc_nt should use sysconfig.get_config_h_filename()
In Python 3.13, we're moving the location of `pyconfig.h` on Windows (in order to support free threaded builds, it now has to be generated/updated at build time, so it can't be static).

To correctly find the file location, we should call `sysconfig.get_config_h_filename()` rather than assuming it'll be in the `PC\` directory.

zooba:
I've used the minimal change here. It would be totally valid to always make this call and only add the path if it differs from `{prefix}\include`, but it still only affects the case where we're in a CPython build directory so I don't see any reason to touch code outside of the checks for that.

Fixes #219

### DIFF
--- a/distutils/sysconfig.py
+++ b/distutils/sysconfig.py
@@ -195,12 +195,11 @@ def _get_python_inc_posix_prefix(prefix):
 
 def _get_python_inc_nt(prefix, spec_prefix, plat_specific):
     if python_build:
-        # Include both the include and PC dir to ensure we can find
-        # pyconfig.h
+        # Include both include dirs to ensure we can find pyconfig.h
         return (
             os.path.join(prefix, "include")
             + os.path.pathsep
-            + os.path.join(prefix, "PC")
+            + os.path.dirname(sysconfig.get_config_h_filename())
         )
     return os.path.join(prefix, "include")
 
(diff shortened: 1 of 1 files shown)


zooba:
Those failures seem to be due to `setuptools` not being installed by default in 3.12. I'd suggest that may indicate that earlier versions are testing the wrong distutils/setuptools versions, or at least missing the opportunity to test with the current code.

The tests impacted by this change passed. It's unfortunate that it aborts the parallel jobs so we can't see their results.

lazka:
lgtm, but there is one more place where the old path is used: https://github.com/pypa/distutils/blob/35069adc37879439205b89165695af62da6edb16/distutils/sysconfig.py#L362-L371

lazka:
> The tests impacted by this change passed. It's unfortunate that it aborts the parallel jobs so we can't see their results.

I've created #222 for this

zooba:
> there is one more place where the old path is used

Is there any reason why this shouldn't just call `sysconfig` all the time? (Or just import the function from sysconfig instead of redefining it?)

lazka:
lgtm

FFY00 (approved):

zooba:
Ping on this issue.

zooba:
Thanks Jason!

Issue #142, opened 2022-05-16
[FR] Support os.PathLike objects


taralx:
I prefer to use pathlib.Path objects, but helpers like `distutils.Command.make_file` only accept strings: https://github.com/pypa/distutils/blob/27638e388d8ae92368e5459254b59a026e73b1cd/distutils/cmd.py#L387-L391

Filing this issue before I go in and make a PR to change these to `os.PathLike` checks. Any issues with that approach?

Directed here from setuptools: https://github.com/pypa/setuptools/issues/3315

jaraco:
Yes, I agree it would be preferable to support Pathlike objects natively. Where possible, I'd like for distutils to simply pass these values along rather than handle them explicitly. Feel free to prepare a PR.

taralx:
Do you mean that I should just remove the string checks?

jaraco:
Yes, I think so. See https://github.com/pypa/distutils/tree/refactor/142-allow-pathlike for how I might handle the situation.

taralx:
Seems like your change is better than what I would have done. I'll take it and see where else the pattern should apply.

PR #230, opened 2024-02-29
ENH: Extension should be able to accept PathLike sources objects

MridulS:
In the spirit of https://github.com/pypa/distutils/issues/142 I think we should be able to pass a list of file objects to the extension class too.

Let me know if there is any else needed here to push this forward :)


jaraco:
I think I got ahead of myself with the merge. I see now the changes are causing issues with both Windows and setuptools integration tests, both of which are expecting string objects for `.sources`, so before a change like this can be accepted, the downstream consumers are all going to need to be able to support `PathLike` objects.

I've gone ahead and force-pushed main back to 08699b63, backing out this change, and I've pushed 9fc96714 to the `pathlike_ext` branch and opened a new PR with that branch to work out the details.

PR #237, opened 2024-03-03
ENH: Extension should be able to accept PathLike sources objects

jaraco:
- **ENH: Extension should be able to accept PathLike sources objects**
- **Adjust expectation to match behavior.**
- **For consistency, ensure Extension.sources is always a path.Path object.**

This PR is a resubmit of #230.


jaraco:
I've managed to fix the Windows tests and I started looking into the setuptools integration test failures. As it turns out, those are going to be harder to solve. I did start [this branch](https://github.com/pypa/setuptools/tree/feature/pathlib-sources-support) to track some of the changes needed, but unfortunately, there's some [pretty nasty code](https://github.com/pypa/setuptools/blob/8c45d6e445a8ca5f5a1fd724a80a5c418fe36780/setuptools/command/egg_info.py#L505-L530) affecting paths. I'm thinking we're at least going to have to get Setuptools to the level of requiring `str` (and not `bytes`) for all paths before attempting to support `pathlib.Path` objects. Trying to support `bytes` and `str` and `pathlib.Path` is just going to be too complicated.

All of this makes me think we may want to defer this work until we can get distutils properly embedded within setuptools, so we don't have to deal with these concerns across two repositories.

lazka:
A more compatible approach would be to just call `os.fspath` on all inputs, and keep the internal paths type as is.

jaraco:
> A more compatible and more limited approach would be to just call `os.fspath` on all inputs, and keep the validation, the internal and exposed paths type as is.

That's a good suggestion. It would be incrementally better and would give users a better experience.

PR #239, opened 2024-03-10
Remove extra quotes from litteral strings

DimitriPapadopoulos:
Automated formatting runs such as #149 / ab77f7d left over these extra quotes.


jaraco:
Thank you!

Issue #240, opened 2024-03-10
Apply Scientific Python repo-review suggestions


DimitriPapadopoulos:
Apply those rules that make sense:
https://learn.scientific-python.org/development/guides/repo-review/?repo=pypa%2Fdistutils&branch=main

Most if not all of these rules already appear in https://github.com/jaraco/skeleton/issues/112. In the issue, I am interested in prerequisites to applying these rules, such as fixing linter issues:

**[RF101](https://learn.scientific-python.org/development/guides/style#RF101): Bugbear must be selected**
Must select the flake8-bugbear B checks. Recommended:
```yaml
[tool.ruff.lint]
extend-select = [
  "B",  # flake8-bugbear
]
```

**[RF102](https://learn.scientific-python.org/development/guides/style#RF102): isort must be selected**
Must select the isort I checks. Recommended:
```yaml
[tool.ruff.lint]
extend-select = [
  "I",  # isort
]
```

**[RF103](https://learn.scientific-python.org/development/guides/style#RF103): pyupgrade must be selected**
Must select the pyupgrade UP checks. Recommended:
```yaml
[tool.ruff.lint]
extend-select = [
  "UP",  # pyupgrade
]
```

PR #243, opened 2024-03-10
Apply ruff/bugbear rules (B)

DimitriPapadopoulos:
Start addressing #240.
```
B006 Do not use mutable data structures for argument defaults
````    
https://docs.astral.sh/ruff/rules/mutable-argument-default/

Don't apply this rule:
````    
B904 Within an `except` clause, raise exceptions with `raise ... from err` or `raise ... from None` to distinguish them from errors in exception handling
````    
https://docs.astral.sh/ruff/rules/raise-without-from-inside-except/

### DIFF
--- a/distutils/command/config.py
+++ b/distutils/command/config.py
@@ -328,7 +328,7 @@ class config(Command):
         library_dirs=None,
         headers=None,
         include_dirs=None,
-        other_libraries: Sequence[str] = [],
+        other_libraries: Sequence[str] = (),
     ):
         """Determine if 'library' is available to be linked against,
         without actually checking that any particular symbols are provided
--- a/distutils/fancy_getopt.py
+++ b/distutils/fancy_getopt.py
@@ -449,7 +449,7 @@ class OptionDummy:
     """Dummy class just used as a place to hold command-line option
     values as instance attributes."""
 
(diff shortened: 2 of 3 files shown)

### CODE REVIEW

**Comment by jaraco** [comment]
Thanks for this request. I'd not seen it before I addressed it in 448a2a12848ca7e99b83958f59db44bb68f6120b. It's okay to pass mutable arguments as long as there's a type declaration protecting against mutating it (which there is now), so the B006 checks no longer fail:

```
 distutils main @ ruff check --select B006
All checks passed!
```

**Comment by DimitriPapadopoulos** [comment]
Nevertheless, it's perhaps better to use `=()` instead of `=[]` as a default value and be consistent about it:
```console
$ fgrep -R '=())'
distutils/tests/test_cmd.py:            cmd.make_file(infiles=1, outfile='', func='func', args=())
distutils/tests/test_cmd.py:        cmd.make_file(infiles='in', outfile='out', func='func', args=())
distutils/tests/test_dist.py:    def create_distribution(self, configfiles=()):
$ 
$ fgrep -R '= [])'
distutils/fancy_getopt.py:    def __init__(self, options: Sequence[Any] = []):
distutils/tests/__init__.py:def missing_compiler_executable(cmd_names: Sequence[str] = []):  # pragma: no cover
$ 
```

PR #244, opened 2024-04-11
Canonicalize fullname, honoring PEP 625

jaraco:
Ref pypa/setuptools#3593 and pypa/setuptools#4286.

- **Support PEP 625**
- **Expect to find canonicalize_* functions in packaging.**
- **Update tests to match new expectation.**
- **Fix canonicalization**
- **In test_sdist, provide a more complex name to capture canonicalization behavior.**
- **Add packaging as a vendored package.**
- **Use vendored packaging.**


distutils/dist.py
```diff
@@ -1189,7 +1191,10 @@ def get_version(self):
         return self.version or "0.0.0"
 
     def get_fullname(self):
-        return f"{self.get_name()}-{self.get_version()}"
+        return "{}-{}".format(
+            canonicalize_name(self.get_name()).replace('-', '_'),
+            canonicalize_version(self.get_version()),
```
abravalheri, line 1196:
```suggestion
            self.get_version(),
```
To avoid problems with trailing zeros as in https://github.com/pypa/setuptools/commit/df45427cbb67c1149fcf5d2d1e2705e69b3baf0c

distutils/dist.py
```diff
@@ -1189,7 +1191,10 @@ def get_version(self):
         return self.version or "0.0.0"
 
     def get_fullname(self):
-        return f"{self.get_name()}-{self.get_version()}"
+        return "{}-{}".format(
+            canonicalize_name(self.get_name()).replace('-', '_'),
+            self.get_version(),
```
abravalheri, line 1196: Maybe something  like https://github.com/pypa/setuptools/pull/4434

jaraco: Thanks for that! Done in 57cab7118.

PR #245, opened 2024-04-13
Apply ruff/refurb rules (FURB)

DimitriPapadopoulos:


distutils/dir_util.py
```diff
@@ -95,8 +95,7 @@ def create_tree(base_dir, files, mode=0o777, verbose=1, dry_run=0):
     """
     # First get the list of directories to create
     need_dir = set()
-    for file in files:
-        need_dir.add(os.path.join(base_dir, os.path.dirname(file)))
+    need_dir.update(os.path.join(base_dir, os.path.dirname(file)) for file in files)
```
jaraco, line 98: Looks like the delinter only removed half the lint. They replaced a init/for/append loop with an init/update operation. Just init would have been better. I'll fix that (can't fix here due to the unmodified line).

Issue #246, opened 2024-04-13
[BUG] setuptools 69.4.0 breaks mpi4py installation


dalcinl:
### setuptools version

69.4.0

### Python version

all

### OS

POSIX

### Description

Originally reported at https://github.com/mpi4py/mpi4py/issues/484
Also reported here for another project: https://gitlab.com/petsc/petsc/-/issues/1579

The commit https://github.com/pypa/setuptools/commit/a131f83e2967514e2973fb36f2ca64e3ac8efc3c changed the return type of a publich method within a public class within a public module.

### Expected behavior

`pip install mpi4py` should succeed with no traceback.

### How to Reproduce

`pip install mpi4py` should be enough.

However, it cannot be reproduced easily as setuptools 69.4.0 was yanked for other reasons.

### Output

```console
File "/tmp/pip-build-env-qxqtzla9/overlay/lib/python3.10/site-packages/setuptools/_distutils/ccompiler.py", line 1232, in gen_lib_options
          lib_opts.extend(always_iterable(compiler.runtime_library_dir_option(dir)))
        File "/tmp/pip-install-u2viu3lx/mpi4py_70e5ecd7d41d4c9fb50ac4e92d5eb59f/conf/mpidistutils.py", line 68, in rpath_option
          if option.startswith('-R'):
      AttributeError: 'list' object has no attribute 'startswith'
```

jaraco:
> changed the return type of a public method within a public class within a public module

Although Python conventions have become more rigorous in the past couple of decades, distutils is ancient and doesn't always conform to modern expectations about public/private visibility (FYI).

That said, you've also observed that this function previously did return strings or lists, and that was why in the review I didn't challenge the change of return type.



> Is there any change setuptools would accept [a change back to a single string]?
 
Yes, maybe. My preference would be to return richer objects that more accurately reflect the semantic meaning of the operation. If `--enable-new-dtags` and `-rpath` are independent settings, I'd prefer for them to be specified independently and not bundled into a single string, especially if common usage is to supply them separately.

That said, I'd like not to be breaking our most popular downstream projects.

Here's what I propose:

- patch distutils/setuptools to restore the prior expectation with a comment about why the change
- with mpi4py and other affected projects, register and issue and add support for lists from this method
- after a certain period (6-12 months?), start relying on a new expectation that lists might be returned

dalcinl:
[...]

PR #247, opened 2024-04-13
Wrap linker arg handling to restore prior expectation.

jaraco:
Closes pypa/distutils#246.

### DIFF
--- a/distutils/compat/__init__.py
+++ b/distutils/compat/__init__.py
@@ -7,7 +7,7 @@ def consolidate_linker_args(args: list[str]) -> str:
     """
     Ensure the return value is a string for backward compatibility.
 
-    Retain until at least 2024-10-31.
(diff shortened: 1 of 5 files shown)


distutils/compat/__init__.py
```diff
@@ -0,0 +1,15 @@
+from __future__ import annotations
+
+from .py38 import removeprefix
+
+
+def consolidate_linker_args(args: list[str]) -> str:
+    """
+    Ensure the return value is a string for backward compatibility.
+
+    Retain until at least 2024-04-31. See pypa/distutils#246
```
dalcinl, line 10: @jaraco Just to double checking... do you really meant 2024-04-31 here? That's about just two weeks from today, not six months!

jaraco: I meant 2025. Fixed in e5268b516.

PR #248, opened 2024-04-19
Extend Windows system executable extensions

LarrysGIT:
For Windows system, the `find_executable` function is hardcoded with `.exe` to locate the executable file.
Windows has a list of execution extensions, which tells by the environment variable `PATHEXT`.
The pull request is aiming to fix the following use case,

```
>>> find_executable("apktool")
*Returns None*
```
The pull request aims to locate the first matched executable
```
>>> find_executable("apktool")
'C:\\bin\\apktool.bat'
```

### DIFF
--- a/distutils/spawn.py
+++ b/distutils/spawn.py
@@ -110,7 +110,9 @@ def find_executable(executable, path=None):
     if os.path.isfile(executable):
         return executable
 
-    for p in _search_paths(path):
-        for exe in filter(pathlib.Path.is_file, _executable_candidates(p / executable)):
-            return os.fspath(exe)
-    return None
+    found = (
+        os.fspath(exe)
+        for p in _search_paths(path)
+        for exe in filter(pathlib.Path.is_file, _executable_candidates(p / executable))
+    )
+    return next(found, None)
--- a/distutils/spawn.py
+++ b/distutils/spawn.py
@@ -77,9 +77,9 @@ def _executable_candidates(executable: pathlib.Path):
     yield executable
     if platform.system() != 'Windows':
         return
-    exts = os.environ.get('PATHEXT').lower().split(os.pathsep)
-    for ext in filter(executable.suffix.__ne__, exts):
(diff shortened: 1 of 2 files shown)


pfmoore:
Note that `CreateProcess` (and hence `subprocess.run` will not recognise arbitrary extensions (it’s hard coded to handle only exe according to the docs, but bat and com also seem to work[1]. So this change may in fact cause failures in later steps.

[1]: As this is undocumented, though, it may not be reliable.

jaraco:
After doing some more refactoring, I've learned that `find_executable` is very similar to `shutil.which` (similar enough that shutil.which was probably copy-pasted from find_executable). `shutil.which` had evolved PATHEXT searching, so this change brings the distutils behavior more inline with what shutil.which is doing.

My instinct - distutils should deprecate `find_executable` and in `spawn` utilize `shutil.which`.

@LarrysGIT Are you using `distutils.spawn.find_executable`? If so, why not use `shutil.which`?

jaraco:
I've decided since I'm likely to deprecate find_executable that I've backed out this change. I've saved the relevant commits (and some subsequent refactoring) in https://github.com/pypa/distutils/tree/distutils-248-saved.

LarrysGIT:
> After doing some more refactoring, I've learned that `find_executable` is very similar to `shutil.which` (similar enough that shutil.which was probably copy-pasted from find_executable). `shutil.which` had evolved PATHEXT searching, so this change brings the distutils behavior more inline with what shutil.which is doing.
> 
> My instinct - distutils should deprecate `find_executable` and in `spawn` utilize `shutil.which`.
> 
> @LarrysGIT Are you using `distutils.spawn.find_executable`? If so, why not use `shutil.which`?

Hi, I was using some 3rd party python project which invokes `apktool`, I do have it in my `PATH`, however, the 3rd party tool keep failing, so I looked around the `find_executable` code and raised the PR.

PR #254, opened 2024-05-18
Apply ruff rule RUF100

DimitriPapadopoulos:
RUF100 Unused `noqa` directive


ruff.toml
```diff
@@ -1,6 +1,7 @@
 [lint]
 extend-select = [
 	"C901",
+	"RUF100",
```
jaraco, line 4: I'm not sure if I'm prepared to accept the toil that will come from merging conflicts in this file. I'm happy to apply the changes but would like to defer enforcing them until the skeleton can accept the rule.

ruff.toml
```diff
@@ -1,6 +1,7 @@
...
 	"W",
```
jaraco, line 5: One way I'm thinking of dealing with this situation is to create a separate section for local rules:

```suggestion
	"C901",
	"W",

	# local
	"RUF100",
```

jaraco: Let's try that and see where it gets us.

PR #260, opened 2024-06-25
Use a separate build directory for free-threading

brenns10:
Hello!

I work on the [drgn debugger](https://github.com/osandov/drgn), and with the recently added 3.13t builds in the manylinux docker, which enable the `--disable-gil` option, we started getting some build failures when building wheels. It turned out that the standard and free-threading builds of 3.13 both use the same `build_temp` and `build_lib` directories.

Drgn has a bit of a customized build system which allows incremental builds within the `build_temp` directory, so the object files from 3.13 were getting reused for the free-threaded build, but the ABI is not compatible anymore. The result was an extension that segfaulted immediately on import.

It seems that a standard distutils extension doesn't have incremental build support, so maybe overlapping `build_temp` and `build_lib` directories isn't a huge problem. But looking through the code, it seems like distutils tries to avoid overlapping these `build_temp` and `build_lib` directories where the compiler could get confused or produce incorrect results. So I thought this might be a suitable fix for distutils itself, rather than just in drgn's build system.

Thanks!

### DIFF
--- a/distutils/tests/test_build.py
+++ b/distutils/tests/test_build.py
@@ -4,6 +4,7 @@ import os
 import sys
 from distutils.command.build import build
 from distutils.tests import support
+from sysconfig import get_config_var
 from sysconfig import get_platform
 
 
@@ -24,6 +25,8 @@ class TestBuild(support.TempdirManager):
         # examples:
         #   build/lib.macosx-10.3-i386-cpython39
         plat_spec = f'.{cmd.plat_name}-{sys.implementation.cache_tag}'
+        if get_config_var('Py_GIL_DISABLED'):
+            plat_spec += 't'
         if hasattr(sys, 'gettotalrefcount'):
             assert cmd.build_platlib.endswith('-pydebug')
             plat_spec += '-pydebug'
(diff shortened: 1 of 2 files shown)


distutils/tests/test_build.py
```diff
@@ -24,6 +25,8 @@ def test_finalize_options(self):
         # examples:
         #   build/lib.macosx-10.3-i386-cpython39
         plat_spec = f'.{cmd.plat_name}-{sys.implementation.cache_tag}'
+        if get_config_var('Py_GIL_DISABLED'):
+            plat_spec += 't'
```
jaraco, line 29: I confirmed that by adding this test (without the fix), the tests now fail, and tests pass with the fix. Nice.

jaraco (approved):

jaraco:
I'm ignoring the diffcov failure - it only checks coverage on a single environment, so until we have a test suite that implements coverage across all supported environments and we have CI environments running with Python free-threading, these lines are going to be uncovered.

brenns10:
Thank you for the quick review!

Issue #263, opened 2024-06-28
Rewrite the commits


jaraco:
Gah. I spent the last couple of days merging PRs in distutils and had forgotten that the issue references needed to be made absolute. Now we've got a bunch of bare numbers that when merged into Setuptools will create nuisance links to the wrong issues.

It does sound like it's possible to reword commits, even merges, with git, so I'm going to try to rewrite a lot of history.

jaraco:
The last merge to setuptools was 55982565e, so I'll be examining the commits since then.

jaraco:
I issued the command: `git rebase -i 5598256 --rebase-merges`, which produced [this TODO file](https://gist.github.com/jaraco/3bcecc1d2685a4f93ae7f92fc196027f).

I then ran a substitution on it: `s/merge -C (.* pull request) #/merge -c \1 #/`. The lowercase `-c` signals to reword the commit.

Git started rewriting the commits allowing me to re-word the merge commits, but then ran into a merge conflict:

```
 distutils main @ git rebase -i 55982565e --rebase-merges
[detached HEAD 48bf47c5c] Merge pull request pypa/distutils#245 from DimitriPapadopoulos/FURB
 Date: Sat Apr 13 09:32:32 2024 -0400
[detached HEAD c6fad4c77] Merge pull request pypa/distutils#247 from pypa/hotfix/246-linker-args-list
 Date: Sat Apr 13 16:40:32 2024 -0400
[detached HEAD 5aff9d862] Merge pull request pypa/distutils#239 from DimitriPapadopoulos/extra_pairs_of_quotes
 Date: Sat Apr 13 20:16:59 2024 -0400
Auto-merging pyproject.toml
CONFLICT (content): Merge conflict in pyproject.toml
Could not apply 851e71af1... Merge-https-github-com-jaraco-skeleton # Merge https://github.com/jaraco/skeleton
```

That doesn't make any sense. No content has changed. How could there possibly be a merge conflict?

jaraco:
I asked Gemini and [it had some ideas](https://gist.github.com/jaraco/3bcecc1d2685a4f93ae7f92fc196027f/1a4bfaadbd92a34226521f4ec8bf6547c8d41338). It seems that maybe `rerere` might be what I'm seeking. Let's try it.

jaraco:
It's not obvious from the docs if `rerere` needs to be enabled when the conflict was originally resolved, in which case it won't be any help here since I resolved the conflict months ago.

[...]

PR #265, opened 2024-06-29
Enforce ruff/Perflint rules (PERF)

DimitriPapadopoulos:
I have not applied this one:
```
PERF203 `try`-`except` within a loop incurs performance overhead
```
I believe the rest of the changes result in more readable code, but I am not certain about this one. In the absence of measured performance gains, I have left it out. I am happy to apply the [`PERF203`](https://docs.astral.sh/ruff/rules/try-except-in-loop/) rule as well if you want me too.

### DIFF
--- a/ruff.toml
+++ b/ruff.toml
@@ -14,7 +14,9 @@ extend-select = [
 	"UP",
 ]
 ignore = [
+	# local
 	"PERF203",
+
 	# https://docs.astral.sh/ruff/formatter/#conflicting-lint-rules
 	"W191",
 	"E111",
--- a/ruff.toml
+++ b/ruff.toml
@@ -8,11 +8,13 @@ extend-select = [
 	"B",
 	"I",
 	"ISC",
+	"PERF",
 	"RUF010",
 	"RUF100",
 	"UP",
 ]
 ignore = [
+	"PERF203",
 	# https://docs.astral.sh/ruff/formatter/#conflicting-lint-rules
 	"W191",
 	"E111",
--- a/distutils/archive_util.py
+++ b/distutils/archive_util.py
@@ -266,8 +266,7 @@ def make_archive(
         raise ValueError(f"unknown archive format '{format}'")
 
     func = format_info[0]
-    for arg, val in format_info[1]:
-        kwargs[arg] = val
+    kwargs.update(format_info[1])
 
     if format != 'zip':
         kwargs['owner'] = owner
--- a/distutils/command/build_ext.py
+++ b/distutils/command/build_ext.py
@@ -638,8 +638,7 @@ class build_ext(Command):
 
         # Do not override commandline arguments
         if not self.swig_opts:
-            for o in extension.swig_opts:
-                swig_cmd.append(o)
+            swig_cmd.extend(extension.swig_opts)
 
         for source in swig_sources:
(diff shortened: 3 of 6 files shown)


distutils/command/sdist.py
```diff
@@ -24,10 +24,10 @@ def show_formats():
     from ..archive_util import ARCHIVE_FORMATS
     from ..fancy_getopt import FancyGetopt
 
-    formats = []
-    for format in ARCHIVE_FORMATS.keys():
-        formats.append(("formats=" + format, None, ARCHIVE_FORMATS[format][2]))
-    formats.sort()
+    formats = [
+        ("formats=" + format, None, ARCHIVE_FORMATS[format][2])
+        for format in sorted(ARCHIVE_FORMATS.keys())
```
jaraco, line 29:
```suggestion
        for format in sorted(ARCHIVE_FORMATS)
```

distutils/unixccompiler.py
```diff
@@ -144,6 +144,9 @@ class UnixCCompiler(CCompiler):
     xcode_stub_lib_format = dylib_lib_format
     if sys.platform == "cygwin":
         exe_extension = ".exe"
+        shared_lib_extension = ".dll.a"
+        dylib_lib_extension = ".dll"
+        dylib_lib_format = "cyg%s%s"
```
jaraco, line 149: This looks like an artifact of something merged from another review... possibly brought about by my force-push to main in #263. I'll want to double-check the diff and tree.

ruff.toml
```diff
@@ -5,11 +5,13 @@ extend-select = [
 
 	# local
 	"ISC",
+	"PERF",
 	"RUF010",
 	"RUF100",
 	"UP",
 ]
 ignore = [
+	"PERF203",
```
jaraco, line 14: This exclusion needs to go below a `# local` indicator. Otherwise, it conflicts with the other ignores from jaraco/skeleton.

distutils/unixccompiler.py
jaraco: I couldn't tell why, but there were three commits near the top that seem to have been rebased. I'm surprised they show up in the diff. In any case, after rebasing the 5 commits onto the current main, these unrelated changes disappear.

ruff.toml
DimitriPapadopoulos: Where is this documented? I have read [Config file discovery](https://docs.astral.sh/ruff/configuration/#config-file-discovery) and other parts of the documentation, but couldn't find anything using _indicator_ and _local_.

DimitriPapadopoulos: By the way, isn't the situation the same in setuptools?
```
        "YTT", # flake8-2020
]
ignore = [
        "TRY301", # raise-within-try, it's handy
        "UP015", # redundant-open-modes, explicit is preferred
        "UP030", # temporarily disabled
        "UP031", # temporarily disabled
        "UP032", # temporarily disabled
        "UP038", # Using `X | Y` in `isinstance` call is slower and more verbose https://github.com/astral-sh/ruff/issues/7871
```

DimitriPapadopoulos:
About the previous failure: my "fix" for `PERF403` was completely wrong. Tests pass now.

jaraco (approved):

PR #266, opened 2024-06-29
 Enforce ruff/tryceratops rules (TRY)

DimitriPapadopoulos:
I have not applied this one:
```
TRY003 Avoid specifying long messages outside the exception class
```
I believe it changes too much code, so I have left it out.  I am happy to apply the [`TRY003`](https://docs.astral.sh/ruff/rules/raise-vanilla-args/) rule as well if you want me too.


distutils/extension.py
```diff
@@ -105,9 +105,9 @@ def __init__(
         **kw,  # To catch unknown keywords
     ):
         if not isinstance(name, str):
-            raise AssertionError("'name' must be a string")
+            raise TypeError("'name' must be a string")
```
DimitriPapadopoulos, line 108: This is a breaking change.  However, I understand the `AssertionError` was supposed to be a catastrophic event any way. Raising a different exception is also a catastrophic event. It is slightly different, but then a catastrophe is a catastrophe.

jaraco: As you can see from other cases that changed in this diff, there were places that AssertionError was explicitly caught. I think your instinct is right here, but it could cause breakage.

I suspect these were originally `assert` statements that were converted to `raise AssertionError` for correctness under python -O (which no one uses).

distutils/file_util.py
```diff
@@ -140,12 +140,13 @@ def copy_file(  # noqa: C901
...
             try:
                 os.link(src, dst)
-                return (dst, 1)
             except OSError:
                 # If hard linking fails, fall back on copying file
                 # (some special filesystems don't support hard linking
                 #  even under Unix, see issue #8876).
                 pass
+            else:
+                return (dst, 1)
```
jaraco, line 149: Honestly, I feel like TRY is overreaching here. There's no way that `return (dst, 1)` is going to cause an OSError, and re-writing with the else clause just distracts from the typical control flow.

DimitriPapadopoulos: In general, this rule might help identify cases where it's not clear which part of the code may throw. The usefulness of the `TRY300` rule depends on the context: number of lines and complexity of the code under `try`.

In this specific case, it is quite clear that the `return` line won't raise en exception. On the other hand, I do not find the `else` clause less readable. The typical control flow is important of course, but so is clarity about what is tested under `try`. Additionally, the error control flow feels important too, enough to have 3 lines of comments.

distutils/extension.py
DimitriPapadopoulos: Agreed.

PR #267, opened 2024-07-04
sysconfig: skip customize_compiler() with MSVC Python again

lazka:
By enabling customize_compiler() when using the mingw compiler class
in https://github.com/pypa/distutils/commit/2ad8784dfeb816829995613fb5fd9818f3e88922 this also enabled it for when the mingw compiler
class was used with a MSVC built CPython.

MSVC CPython doesn't have any of the config vars that are required in
customize_compiler() though. And while it would be nice if all the env
vars would be considered in that scenario too, it's not clear how this
should be implemented without any sysconfig provided fallbacks
(if CC isn't set but CFLAGS is, there is no way to pass things to
set_executables() etc.)

Given that, just restore the previous behaviour, skip customize_compiler()
with MSVC Python in all cases, and add a test.

Fixes https://github.com/pypa/distutils/issues/268

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -143,6 +143,9 @@ jobs:
             git
       - name: Install Dependencies
         shell: msys2 {0}
+        env:
+          CC: ${{ matrix.cc }}
+          CXX: ${{ matrix.cxx }}
         run: |
           export VIRTUALENV_NO_SETUPTOOLS=1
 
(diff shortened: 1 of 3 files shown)


lazka:
(the first commit is a CI build fix, which was caused by setuptools on pypi now behaving differently)

jaraco:
Re-merged as e221581f8 (I forgot to make the pr reference project-absolute).

lazka:
Thanks!

PR #270, opened 2024-07-12
Fix cygwin CI job

### LINKED ISSUE #269: Two test failures in Cygwin
In 1bcd839f1, I restored the running of cygwin tests for distutils after I realized that the "rust compiler" limitation only needed to affect the "pytest-ruff" dependency, now excluded upstream.

However, re-enabling the tests revealed we've had a couple of regressions since tests were disabled.

```
________ TestMingw32CCompiler.test_customize_compiler_with_msvc_python _________

self = <distutils.tests.test_mingwccompiler.TestMingw32CCompiler object at 0x6ffffe1d04c0>

    def test_customize_compiler_with_msvc_python(self):
        from distutils.cygwinccompiler import Mingw32CCompiler
    
        # In case we have an MSVC Python build, but still want to use
        # Mingw32CCompiler, then customize_compiler() shouldn't fail at least.
        # https://github.com/pypa/setuptools/issues/4456
>       compiler = Mingw32CCompiler()

distutils/tests/test_mingwccompiler.py:54: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

self = <distutils.cygwinccompiler.Mingw32CCompiler object at 0x6ffffde848b0>
verbose = False, dry_run = False, force = False

    def __init__(self, verbose=False, dry_run=False, force=False):
        super().__init__(verbose, dry_run, force)
    
        shared_option = "-shared"
    
        if is_cygwincc(self.cc):
>           raise CCompilerError('Cygwin gcc cannot be used with --compiler=mingw32')
E           distutils.errors.CCompilerError: Cygwin gcc cannot be used with --compiler=mingw32

distutils/cygwinccompiler.py:274: CCompilerError
______ [doctest] distutils.unixccompiler.UnixCCompiler.find_library_file _______
371         data to go on: GCC seems to prefer the shared library, so
372         assume that *all* Unix C compilers do,
373         ignoring even GCC's "-static" option.
374 
375         >>> compiler = UnixCCompiler()
376         >>> compiler._library_root = lambda dir: dir
377         >>> monkeypatch = getfixture('monkeypatch')
378         >>> monkeypatch.setattr(os.path, 'exists', lambda d: 'existing' in d)
379         >>> dirs = ('/foo/bar/missing', '/foo/bar/existing')
380         >>> compiler.find_library_file(dirs, 'abc').replace('\\', '/')
Expected:
    '/foo/bar/existing/libabc.dylib'
Got:
    '/foo/bar/existing/cygabc.dll'

/cygdrive/d/a/distutils/distutils/distutils/unixccompiler.py:380: DocTestFailure
```

@lazka Can you take a look and suggest fixes?

lazka:
See the commits for details.

Feedback welcome.

Fixes #269

### DIFF
--- a/distutils/tests/test_mingwccompiler.py
+++ b/distutils/tests/test_mingwccompiler.py
@@ -45,6 +45,7 @@ class TestMingw32CCompiler:
         with pytest.raises(CCompilerError):
             distutils.cygwinccompiler.Mingw32CCompiler()
 
+    @pytest.mark.skipif('sys.platform == "cygwin"')
     def test_customize_compiler_with_msvc_python(self):
         from distutils.cygwinccompiler import Mingw32CCompiler
 
--- a/distutils/tests/test_unixccompiler.py
+++ b/distutils/tests/test_unixccompiler.py
@@ -315,3 +315,33 @@ class TestUnixCCompiler(support.TempdirManager):
         self.cc.output_dir = 'scratch'
         os.chdir(self.mkdtemp())
         self.cc.has_function('abort')
+
+    def test_find_library_file(self, monkeypatch):
+        compiler = UnixCCompiler()
+        compiler._library_root = lambda dir: dir
+        monkeypatch.setattr(os.path, 'exists', lambda d: 'existing' in d)
+
+        dylibname = 'libabc.dylib' if sys.platform != 'cygwin' else 'cygabc.dll'
+        dirs = ('/foo/bar/missing', '/foo/bar/existing')
+        assert (
+            compiler.find_library_file(dirs, 'abc').replace('\\', '/')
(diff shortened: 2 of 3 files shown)


DWesl:
That's a sensible way to deal with the doctest failures.

jaraco:
Thanks for this. I'm inclined to keep the doctests. They are borderline too complicated for doctests, but they do provide some value in illustrating to the reader what the function does. I do agree it's getting a little convoluted, so fine to port it to another test. Thanks for the help!

lazka:
Thanks!

PR #271, opened 2024-07-19
Allow path objects for data-files

jaraco:
Ported from pypa/setuptools#4495 by @InvincibleRMC.

### DIFF
--- a/distutils/command/install_data.py
+++ b/distutils/command/install_data.py
@@ -5,11 +5,17 @@ platform-independent data files."""
 
 # contributed by Bastian Kleineidam
 
+import functools
 import os
 import pathlib
 
+from typing import Tuple, Iterable
+
 from ..core import Command
 from ..util import change_root, convert_path
+
+
+StrPath = str | pathlib.Path
 
 
 class install_data(Command):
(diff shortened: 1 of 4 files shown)


distutils/command/install_data.py
```diff
@@ -56,6 +57,16 @@ def run(self):
                     )
                 (out, _) = self.copy_file(f, self.install_dir)
                 self.outfiles.append(out)
+            elif isinstance(f, Path):
```
jaraco, line 60: This covers the case that one of the data files is a simple pathlib.Path, but it doesn't cover the case if a `pathlib.Path` is used in a tuple (in the else block below).

InvincibleRMC: Ah you are right. Should I fix it or do you want to fix it?

jaraco: I'm on it. But thanks for offering.

jaraco: My thinking is that `convert_path` already wraps all the places where a Path might be encountered, so probably that should be expanded to account for `pathlib.Path`.

jaraco:
Looks like I jumped the gun on the merge. I've backed out the merge and going to re-submit.

PR #273, opened 2024-07-21
venv_install_options: add missing clear_argv fixture

lazka:
Otherwise the test fails if arguments are passed to pytest, for example --no-cov:

FAILED distutils/tests/test_dist.py::TestDistributionBehavior::test_venv_install_options -
      distutils.errors.DistutilsArgError: option --no-cov not recognized


lazka:
Thanks!

PR #278, opened 2024-08-11
handle failures to find a user home directory

cj81499:
In certain environments (notably, the [bazel sandbox](https://bazel.build/docs/sandboxing) on windows), it is possible for `pathlib.Path('~').expanduser()` to fail to find the user home directory and [raise a `RuntimeError`](https://docs.python.org/3/library/pathlib.html#pathlib.Path.expanduser). This causes `distutils` (and ultimately, `setuptools`) to fail.

With this patch, we catch and handle the exception by logging a warning and continuing without a user's config file.

motivated by https://github.com/bazelbuild/rules_python/pull/1067


distutils/dist.py
```diff
@@ -354,7 +354,12 @@ def _gen_paths(self):
         prefix = '.' * (os.name == 'posix')
         filename = prefix + 'pydistutils.cfg'
         if self.want_user_cfg:
-            yield pathlib.Path('~').expanduser() / filename
+            try:
+                user_home = pathlib.Path('~').expanduser()
+            except RuntimeError:
+                self.announce("Failed to locate user home directory. Skipping user config.", logging.WARNING)
+            else:
+                yield user_home / filename
```
abravalheri, line 362: Hello (disclaimer: not a distutils maintainer).

Would it make sense to use a proper warning here?

```suggestion
            try:
                yield pathlib.Path('~').expanduser() / filename
            except RuntimeError:
                warnings.warn("Failed to locate user home directory. Skipping user config.")
```

cj81499: Maybe! I'm certainly open to other error handling approaches, so long as a failure to locate the home directory doesn't cause failure. Let's see what a maintainer has to say.

jaraco: I don't have an instinct on the matter, so I'd defer to abravalheri's instinct. Let's go with that.

cj81499:
> In other scenarios, I've taken measures to [explicitly set the home dir](https://github.com/jaraco/pytest-home/blob/9a31de5a76aa9e0df54b4a376eb6c831242420bf/pytest_home/fixtures.py#L41-L49) to a known location.

That's _pretty much_ the workaround I implemented in https://github.com/bazelbuild/rules_python/pull/1067/files#diff-a92cc6468f47191b8e22d4bbe2564d3968d3dd4bbc383b9f3df3c4c520e1583eR157-R165.

> Wouldn't it be better for bazel sandbox to provide a suitable (empty) home directory and thus address this concern for any project running in the sandbox?

Personally, I can't think of a reason for `setuptools`/`distutils` to put the burden of ensuring a home directory exists on its users and/or other projects (`rules_python`/`bazel`).
Failing to find a home directory that might contain a user config file and failing to find a user config file within a located home directory are both ways that `setuptools`/`distutils` might fail to find a user config file. I would expect `setuptools`/`distutils` to behave the same (continue the build without using user configuration) in both cases.

For what it's worth, other Python build backends (eg: flit, hatchling) do not fail in the event that you attempt to build without a valid home directory.

jaraco:
> Personally, I can't think of a reason for `setuptools`/`distutils` to put the burden of ensuring a home directory exists on its users and/or other projects (`rules_python`/`bazel`).

I think I agree with you here. However, there's another factor to consider. Do you want `rules_python` to be sensitive to settings in the user's home directory? The reason I created those exclusions was not because the project failed when there was no home directory, but because the user's configuration changed the default behaviors, affecting tests that were meant to run in isolation. My guess is that `rules_python` would want the same isolation, which you'll lose if we roll out this change and then you remove the workaround.

That said, you make a compelling argument. There's not a very good reason for distutils not to be resilient under such conditions.

distutils/dist.py
```diff
@@ -346,7 +346,12 @@ def _gen_paths(self):
         prefix = '.' * (os.name == 'posix')
         filename = prefix + 'pydistutils.cfg'
         if self.want_user_cfg:
-            yield pathlib.Path('~').expanduser() / filename
+            try:
+                yield pathlib.Path('~').expanduser() / filename
+            except RuntimeError:
+                warnings.warn(
```
jaraco, line 352: Diffcov says lines 351 and 352 are uncovered. That's probably not crucial that they're covered, other than someone might come along later and optimize this code and there's nothing protecting the expectation. At the very least, we should have a `# pragma: no cover` comment on line 351 to suppress the issue _and_ signal to a future reviewer that the behavior is uncovered (and thus requires more scrutiny to remove).

```suggestion
            except RuntimeError:  # pragma: no cover
                warnings.warn(
```

distutils/dist.py
```diff
@@ -346,7 +346,12 @@ def _gen_paths(self):
...
+                    "Failed to locate user home directory. Skipping user config."
+                )
```
jaraco, line 354: I'm wondering if we even need a warning. This code is already branchy enough. I'm thinking maybe just suppress the error:

```suggestion
            with contextlib.suppress(RuntimeError):
                yield pathlib.Path('~').expanduser() / filename
```

That will also address the coverage issue.

jaraco (approved):

distutils/dist.py
jaraco: I'm going to go with the `suppress` approach.

cj81499:
> Do you want `rules_python` to be sensitive to settings in the user's home directory?

Personally, no. But my understanding is that bazel's sandbox prevents reads/writes outside of the sandbox (plus, [bazel builds are often performed remotely](https://bazel.build/remote/rbe), in which case the home directory of the user invoking `bazel build <TARGET>` won't be present). In either case, this should now silently fail to find the user's config file (as desired!)

PR #279, opened 2024-08-12
Remove outdated `ImportError` check for `warnings`

abravalheri:
If I am not mistaken `warnings` is a valid import in any supported version of Python, so there is no need for handling an `ImportError`.

This seems to have been introduced in https://github.com/pypa/distutils/commit/c3c5a89898505fe9b1efe3310fc6a666212bd148, 31th Oct 2002 (so the commit might have been accounting for versions of Python prior to [2.1, which was released on 17th Apr 2001](https://docs.python.org/3/whatsnew/2.1.html)).

### DIFF
--- a/distutils/dist.py
+++ b/distutils/dist.py
@@ -14,9 +14,8 @@ import warnings
 from collections.abc import Iterable
 from email import message_from_file
 
-from ._vendor.packaging.utils import canonicalize_name, canonicalize_version
-
 from ._log import log
+from ._vendor.packaging.utils import canonicalize_name, canonicalize_version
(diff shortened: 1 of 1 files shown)


abravalheri:
The change in https://github.com/pypa/distutils/pull/279/commits/8c6015c5d75641ca42d06db5021763c4f12c1d32 is unrelated but seems to be necessary to satisfy the `ruff` linter given the existing configurations.

abravalheri:
The `diffcov` fail seems to be a false positive (the conditional block seems to never have been exercised in tests, so in that regard, not much have changed).

PR #282, opened 2024-08-17
Remove unused wininst-*.exe files

lazka:
They where used by bdist_wininst, which was removed long ago in https://github.com/pypa/distutils/pull/170


lazka:
Unless I'm missing something?

This reduces the setuptools wheel size from 2.3MB to 1.3MB.

PR #285, opened 2024-08-25
Use arm64 MSVC on arm64 Windows

saschanaz:
Related to https://github.com/pypa/setuptools/pull/4555 but with no dependency.

### DIFF
--- a/distutils/_msvccompiler.py
+++ b/distutils/_msvccompiler.py
@@ -181,8 +181,7 @@ def _find_exe(exe, paths=None):
 def _get_plat_to_vcvars():
     # A map keyed by get_platform() return values to values accepted by
     # 'vcvarsall.bat'.
-    host_platform = get_host_platform()
-    if host_platform == get_platform() and host_platform == "win-arm64":
+    if get_platform() == get_host_platform() == "win-arm64":
         # Use the native MSVC host if the host platform would need expensive
         # emulation for x86.
         return {
--- a/distutils/_msvccompiler.py
+++ b/distutils/_msvccompiler.py
@@ -178,26 +178,31 @@ def _find_exe(exe, paths=None):
     return exe
 
 
-# A map keyed by get_platform() return values to values accepted by
-# 'vcvarsall.bat'.
-if get_platform() == get_host_platform() and get_host_platform() == "win-arm64":
-    # Use the native MSVC host if the host platform would need expensive
-    # emulation for x86.
-    PLAT_TO_VCVARS = {
-        'win32': 'arm64_x86',
-        'win-amd64': 'arm64_amd64',
-        'win-arm32': 'arm64_arm',
-        'win-arm64': 'arm64',
-    }
-else:
-    # Always cross-compile from x86 to work with the lighter-weight MSVC
-    # installs that do not include native 64-bit tools.
-    PLAT_TO_VCVARS = {
-        'win32': 'x86',
-        'win-amd64': 'x86_amd64',
-        'win-arm32': 'x86_arm',
-        'win-arm64': 'x86_arm64',
-    }
+def _get_plat_to_vcvars():
+    # A map keyed by get_platform() return values to values accepted by
+    # 'vcvarsall.bat'.
+    host_platform = get_host_platform()
+    if host_platform == get_platform() and get_host_platform() == "win-arm64":
+        # Use the native MSVC host if the host platform would need expensive
+        # emulation for x86.
+        return {
+            'win32': 'arm64_x86',
+            'win-amd64': 'arm64_amd64',
+            'win-arm32': 'arm64_arm',
+            'win-arm64': 'arm64',
+        }
+    else:
+        # Always cross-compile from x86 to work with the lighter-weight MSVC
+        # installs that do not include native 64-bit tools.
+        return {
+            'win32': 'x86',
+            'win-amd64': 'x86_amd64',
+            'win-arm32': 'x86_arm',
+            'win-arm64': 'x86_arm64',
+        }
+
+
+PLAT_TO_VCVARS = _get_plat_to_vcvars()
 
 
 class MSVCCompiler(CCompiler):
--- a/distutils/_msvccompiler.py
+++ b/distutils/_msvccompiler.py
@@ -198,8 +198,8 @@ def _get_vcvars_spec(host_platform, platform):
     >>> _get_vcvars_spec('win-arm64', 'win-amd64')
     'arm64_amd64'
 
-    Always cross-compile from x86 to work with the lighter-weight MSVC
-    installs that do not include native 64-bit tools.
+    Otherwise, always cross-compile from x86 to work with the
+    lighter-weight MSVC installs that do not include native 64-bit tools.
 
     >>> _get_vcvars_spec('win32', 'win32')
     'x86'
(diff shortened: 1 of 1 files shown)


zooba:
An alternative approach would be to conditionally construct `PLAT_TO_VCVARS` based on `get_host_platform` (which would then let you cross-compile to X86 or AMD64 using the native ARM64 compilers on ARM64).

The host platform should not be changing due to settings or anything else at runtime (the target platform theoretically should be able to be configured at runtime, though I'm not sure if it can).

distutils/_msvccompiler.py
```diff
@@ -178,15 +178,30 @@ def _find_exe(exe, paths=None):
...
-PLAT_TO_VCVARS = {
-    'win32': 'x86',
-    'win-amd64': 'x86_amd64',
-    'win-arm32': 'x86_arm',
-    'win-arm64': 'x86_arm64',
-}
+def _get_plat_to_vcvars():
+    # A map keyed by get_platform() return values to values accepted by
+    # 'vcvarsall.bat'.
+    if get_platform() == get_host_platform() == "win-arm64":
```
zooba, line 184:
```suggestion
    if get_host_platform() == "win-arm64":
```

No reason to trust `get_platform` here, and it's used later on to look up the specific result from this dict anyway.

distutils/_msvccompiler.py
```diff
@@ -178,15 +178,30 @@ def _find_exe(exe, paths=None):
...
-# A map keyed by get_platform() return values to values accepted by
-# 'vcvarsall.bat'. Always cross-compile from x86 to work with the
-# lighter-weight MSVC installs that do not include native 64-bit tools.
-PLAT_TO_VCVARS = {
-    'win32': 'x86',
-    'win-amd64': 'x86_amd64',
-    'win-arm32': 'x86_arm',
-    'win-arm64': 'x86_arm64',
-}
+def _get_plat_to_vcvars():
```
zooba, line 181: I don't think anything is really gained from using a function compared to just putting the `if` statements in the code, but I also don't own this module anymore so I'm not going to push back too hard.

distutils/_msvccompiler.py
```diff
@@ -178,15 +178,24 @@ def _find_exe(exe, paths=None):
     return exe
 
 
-# A map keyed by get_platform() return values to values accepted by
```
jaraco, line 181: I noticed we lost this definition of `PLAT_TO_VCVARS`. It might be nice to bring that back somehow, but I'm struggling with how, because it is now defined twice. I'm going to try some things.

saschanaz: For now I restored the comment... but yes IDEs would be confused 🤔

saschanaz: VSCode seems to not sure anything anyway even with the current main branch, huh.

zooba: IDEs have to be able to handle multiple conditional definitions in Python - it's a fundamental part of the language (it's even supported in type stub files). If they don't, file bugs with the IDEs.

distutils/_msvccompiler.py
```diff
@@ -178,17 +178,43 @@ def _find_exe(exe, paths=None):
...
+
+    Uses the native MSVC host if the host platform would need expensive
+    emulation for x86.
+
+    >>> _get_vcvars_spec('win-arm64', 'win32')
+    'arm64_x86'
+    >>> _get_vcvars_spec('win-arm64', 'win-amd64')
+    'arm64_amd64'
+
+    Always cross-compile from x86 to work with the lighter-weight MSVC
```
saschanaz, line 201: "Otherwise always ..." ?

jaraco:
It was really bothering me that we had two copies of essentially the same mapping, so I spent some time thinking about what is it that these two mappings represent and distilled from that a function that will construct the spec. This new function `_get_vcvars_spec` encodes the logic that was previously in comments only (that `win32` supersedes everything but `win-arm64`). It additionally adds some doctests that perform some checks on the expected values. This logic should be more durable and maintainable than lookup tables.

distutils/_msvccompiler.py
jaraco: Done in 761162d7e.

jaraco:
diffcov tests are now failing for the changes in `initialize`, which is due to the fact that the coverage tests run on Linux and this project doesn't have coverage aggregation, so I'll ignore it for now.

Let me know if you have any lingering concerns; otherwise, it's ready to go.

jaraco (approved):

saschanaz:
Confirmed the behavior locally, LGTM

zooba:
> This logic should be more durable and maintainable than lookup tables.

It's a bigger compatibility change from what's currently there, and it seems possible (I won't say likely) that someone could be depending on that dict being there, either to read it or to update it.

Your call, of course, but if you're going to break compatibility then why not just implement it better in setuptools?

jaraco:
> > This logic should be more durable and maintainable than lookup tables.
> 
> It's a bigger compatibility change from what's currently there, and it seems possible (I won't say likely) that someone could be depending on that dict being there, either to read it or to update it.

My compatibility strategy for distutils is to first deprecate all such interactions with distutils (such that only setuptools is interacting with distutils), making distutils an implementation detail that Setuptools can then absorb (accept and run the tests) and converge.

The deprecation is in place, so there shouldn't be any users relying on this variable.

I know there's a possibility users are reliant on it and just haven't admitted such, but I'm willing to take that risk and provide a compatibility layer or back it out if necessary.

> Why not just implement it better in setuptools? (Otherwise you'll never get rid of distutils ;) )

I see much of what Setuptools provides over distutils as smallish extensions of the core distutils functionality. I'm expecting that many (most?) of the interfaces in distutils will become the canonical implementation for Setuptools. I've been consolidating some behaviors and directing contributions here when relevant, rather than expanding the customizations in Setuptools.

distutils/_msvccompiler.py
```diff
@@ -178,28 +178,41 @@ def _find_exe(exe, paths=None):
...
+    >>> _get_vcvars_spec('win-arm32', 'win-arm32')
+    'x86_arm'
+    >>> _get_vcvars_spec('win-amd64', 'win-arm64')
+    'x86_arm64'
+    """
+    if host_platform != 'win-arm64':
+        host_platform = 'win32'
+    vc_hp = _vcvars_names[host_platform]
+    vc_plat = _vcvars_names[platform]
+    return vc_hp if vc_hp == vc_plat else f'{vc_hp}_{vc_plat}'
```
Avasam: @jaraco looks like this specific change may have broken cross-compilation (at least in some contexts). See investigation done in https://github.com/pypa/setuptools/issues/4648#issuecomment-2351152196

distutils/_msvccompiler.py
```diff
@@ -242,13 +268,12 @@ def initialize(self, plat_name=None):
...
-        if plat_name not in PLAT_TO_VCVARS:
+        if plat_name not in _vcvars_names:
             raise DistutilsPlatformError(
-                f"--plat-name must be one of {tuple(PLAT_TO_VCVARS)}"
+                f"--plat-name must be one of {tuple(_vcvars_names)}"
             )
 
-        # Get the vcvarsall.bat spec for the requested platform.
-        plat_spec = PLAT_TO_VCVARS[plat_name]
+        plat_spec = _get_vcvars_spec(get_host_platform(), get_platform())
```
saschanaz, line 276: Seems the bug is in line 276 where it's ignoring the `plat_name` parameter and instead calling `get_platform()` again. 👀

Avasam: I think you're right, the following is enough of a change:
```suggestion
        plat_spec = _get_vcvars_spec(get_host_platform(), plat_name)
```

PR #287, opened 2024-08-27
Remove legacy msvc compiler modules.

jaraco:
They were marked as deprecated in c802880bfba39d3072599eef2e0f34568f93e6cd, so they shouldn't be in use.


alex:
Notwithstanding that these shouldn't be in use, in practice these appear to be used by cffi, and thus this change breaks cffi: https://github.com/python-cffi/cffi/blob/main/src/cffi/_shimmed_dist_utils.py#L33

I'll file a cffi bug about fixing this.

eli-schwartz:
It also broke gnome: https://gitlab.gnome.org/GNOME/gobject-introspection/-/issues/515

illume:
broke pygame

jaraco:
> broke pygame

If you have a reference to the downstream report, that would be helpful to follow. I looked briefly into the other issues and I don't think there's anything this project should do, though I could consider re-adding some interfaces to restore compatibility for a limited duration, but only if there's a compelling case. In general, I expect users reliant on these older modules to be in a precarious state anyhow.

illume:
For pygame the setuptools for windows msvc compilers have been pinned to previous setuptools version. I haven't looked yet on how to update the code which uses the compiler yet. It's here if you're interested: https://github.com/pygame/pygame/blob/main/buildconfig/vstools.py

I know not everyone can pin their packages to previous versions of setuptools. But for anyone that can pin whilst they fix, I wrote instructions here: https://github.com/pypa/setuptools/issues/4620#issuecomment-2323233905

nitzmahone:
Looks like the `cffi` "usage" is an ancient and likely-all-dead monkeypatch- spinning an updated CFFI release now that makes the import/patch optional until we're sure it's "all dead" and can go through its pockets for loose change :wink:.

PR #288, opened 2024-08-27
Remove `_vendor` directory

abravalheri:
Setuptools recently changed its vendoring method to address the criticism regarding the proliferation of `_vendor` directories and multiple copies of the same files being distributed in the same wheel/sdist: https://github.com/pypa/setuptools/issues/4455.

That change however do not address the existence of `setuptools/_distutils/_vendor` and therefore do not close the original issue 100%. I think there is not much point in `distutils` maintaining its own vendoring system, so this PR proposes simply removing it, and relying on the dependencies that come installed with setuptools.

In https://github.com/pypa/setuptools/issues/4594, we also got a report regarding external tools expecting `setuptools/_distutils/_vendor/*.dist-info` folders to behave in a certain way (despite it being a private directory and not added to `sys.path`). Although that issue is not really a bug in setuptools (in my opinion), simply removing `distutils/_vendor` would also help with that.


distutils/dist.py
```diff
@@ -26,6 +25,13 @@
 from .fancy_getopt import FancyGetopt, translate_longopt
 from .util import check_environ, rfc822_escape, strtobool
 
+try:
+    from packaging.utils import canonicalize_name, canonicalize_version
+except ImportError as ex:  # pragma: no cover
+    raise ImportError(
+        "Distutils should be distributed alongside setuptools and its dependencies"
+    ) from ex
```
abravalheri, line 33: This exception is not really necessary, just adds a bit more verbosity to help debugging if necessary. We might as well just remove the `try...except` wrapper and run the import statement directly.

pyproject.toml
```diff
@@ -17,7 +17,7 @@ classifiers = [
 	"Programming Language :: Python :: 3 :: Only",
 ]
 requires-python = ">=3.8"
-dependencies = []
+dependencies = ["packaging"]  # setuptools dependency
```
jaraco, line 20: Let's make the dependencies multi-line, for easier diffs.

Also, I'm unsure about the comment. It communicates that it is a setuptools dependency currently, but that could change. How about instead a comment indicating that setuptools needs to provide all dependencies?

```suggestion
dependencies = [
	# Setuptools must require these
	"packaging",
]
```

Adding this comment pushes me a little closer to releasing distutils as its own package (with a caveat that it's not meant to be installed anywhere), and then including it as a vendored package in Setuptools (just like any other, with its dependencies). I'm not prepared to do that yet, but it's more tempting now that distutils has dependencies.

distutils/dist.py
jaraco: Agreed. Let's just import it naturally.

jaraco: Done in 9d05e00f4.

jaraco:
The diffcov errors are weird, probably there due to the broken main brought about by a nitpicky change in ruff import sorting, which is now fixed in 7ee6a6e42.

abravalheri:
Thank you very much for the review and improvements @jaraco.

> Adding this comment pushes me a little closer to releasing distutils as its own package (with a caveat that it's not meant to be installed anywhere), and then including it as a vendored package in Setuptools (just like any other, with its dependencies). I'm not prepared to do that yet, but it's more tempting now that distutils has dependencies.

There is another approach as well which is to "just let setuptools" do its thing and remove this concern from `distutils` (instead, just do a weaker/simpler canonicalization that does not require 3rd party packages[^1]).
Currently setuptools patches `DistutilsMetadata` because there is no mechanism for using a subclass when `setuptools.dist.Distribution` is instantiated... So for now we are stuck with this approach, and therefore adding dependencies to `distutils` has little pay off.

In the future we either start moving code from setuptools to distutils, or add a method that allows using `DistutilsMetadata` subclasses when instantiating `setuptools.dist.Distribution`[^2].

[^1]: We can safely assume that by the stage that we need the canonicalization, name and version have already been validated, so we could use some regex for the `alpha|beta|rc => a|b|c` and `.|- => _` replacements.

[^2]: For example in https://github.com/pypa/distutils/blob/de0bb446b5a953e8fa48998980c07496533968bc/distutils/dist.py#L149, instead of having `self.metadata = DistributionMetadata()` we could do `self.metadata = self._create_metdata()`, do that we can extend `DistutilsMetadata` via subclassing instead of patching.

PR #289, opened 2024-08-27
Consolidate VS 2017 support from Setuptools

jaraco:
Prerequisite for pypa/setuptools#4600

- **Apply downstream VS 2017 support.**
- **Apply error message from downstream.**


jaraco:
Ignoring diffcov (Windows runs aren't counted).

PR #296, opened 2024-09-09
cygwinccompiler: Get the compilers from sysconfig

naveen521kk:
On the CLANG64 environment of MSYS2, we should use clang
instead of gcc. This patch gets the compiler from sysconfig which
would be set when Python was built.

Without this patch, the build fails when it's trying to check if the
compiler is cygwin's one.

### DIFF
--- a/distutils/cygwinccompiler.py
+++ b/distutils/cygwinccompiler.py
@@ -21,6 +21,7 @@ from .errors import (
     DistutilsPlatformError,
 )
 from .file_util import write_file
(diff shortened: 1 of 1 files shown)


lazka:
lgtm

(There is potential for moving all this to customize_compiler() but we would have to somehow get rid of is_cygwincc(), or only call it when actually using the compiler, as it is currently expects the compiler to be set before customize_compiler(). Not clear, so let's keep things simple)

PR #298, opened 2024-09-14
Fix cross-platform compilation using `distutils._msvccompiler.MSVCCompiler`

Avasam:
Actually use the `plat_name` param in `MSVCCompiler.initialize`
Thanks @saschanaz https://github.com/pypa/distutils/pull/285#discussion_r1759843926

Fixes https://github.com/pypa/setuptools/issues/4648#issuecomment-2351176463

### DIFF
--- a/distutils/tests/test_msvccompiler.py
+++ b/distutils/tests/test_msvccompiler.py
@@ -2,11 +2,13 @@
 
 import os
 import sys
+import sysconfig
 import threading
 import unittest.mock as mock
 from distutils import _msvccompiler
 from distutils.errors import DistutilsPlatformError
 from distutils.tests import support
+from distutils.util import get_platform
 
 import pytest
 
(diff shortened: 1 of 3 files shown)


distutils/tests/test_msvccompiler.py
```diff
@@ -28,6 +30,28 @@ def _find_vcvarsall(plat_spec):
                 'wont find this version',
             )
 
+    @pytest.mark.skipif(
+        not sysconfig.get_platform().startswith("win"),
```
Avasam, line 34: Alternatively:
```suggestion
        sysconfig.get_platform() not in _msvccompiler._vcvars_names,
```

distutils/tests/test_msvccompiler.py
```diff
@@ -28,6 +30,28 @@ def _find_vcvarsall(plat_spec):
...
+    )
+    def test_cross_platform_compilation_paths(self, monkeypatch, plat_name, expected):
+        compiler = _msvccompiler.MSVCCompiler()
+
+        # makes sure that the right target platform name is used
+        def _get_vcvars_spec(host_platform, platform):
+            assert platform == expected
+
+        monkeypatch.setattr(_msvccompiler, '_get_vcvars_spec', _get_vcvars_spec)
+        compiler.initialize(plat_name)
```
jaraco, line 53:
```suggestion
    def test_cross_platform_compilation_paths(self, monkeypatch, plat_name, expected):
        """
        Ensure a specified target platform is passed to _get_vcvars_spec.
        """
        compiler = _msvccompiler.MSVCCompiler()

        def _get_vcvars_spec(host_platform, platform):
            assert platform == expected

        monkeypatch.setattr(_msvccompiler, '_get_vcvars_spec', _get_vcvars_spec)
        compiler.initialize(plat_name)
```

jaraco (approved):

newsfragments/298.bugfix.rst
```diff
@@ -0,0 +1 @@
+Fix cross-platform compilation using ``distutils._msvccompiler.MSVCCompiler`` -- by :user:`saschanaz` and :user:`Avasam`
```
jaraco, line 1: Oh, gosh. I'm realizing now that we can't use these news fragments. They refer to issues in this repository, which is merged into Setuptools, so adding them here will cause setuptools to render them as their own issues. I'll just delete it.

Avasam: Oh right!
Well you could still have newsfragments without issue # if it starts with `+` (for example: `+distutils298.bugfix.rst`)

jaraco: Good idea. I'll aim to do things like that in the future.

Issue #169, opened 2022-08-13
New test_set_include_dirs leaves files in working directory


jaraco:
In #153, I added [test_set_include_dirs](https://github.com/pypa/distutils/blob/cc017c77948737d131f683e0c25cd37bc639b8fc/distutils/tests/test_ccompiler.py#L43-L55). At least on my mac, this operation leaves a file in the working directory:

`private/var/folders/sx/n5gkrgfx6zd91ymxr2sr9wvw00n8zm/T/pytest-of-jaraco/pytest-32/test_set_include_dirs0/foo.o`

It seems something about invoking a compiler on a file causes the path for that file to get built relative to the current directory :/

jaraco:
So apparently, the C compiler class [intentionally](https://github.com/pypa/distutils/blob/cc017c77948737d131f683e0c25cd37bc639b8fc/distutils/ccompiler.py#L929) replaces absolute paths with relative ones.

jaraco:
Removing that line (and the one above it), the tests still pass, so there's no good way to tell how important that behavior is. Do all known use-cases rely on relative paths? Or do some pass absolute paths and rely on this rewriting?

In my opinion, the behavior should be consistent, to place the compiled file alongside its source regardless of whether it was specified absolute or relative. It's going to take some work to avoid potential breakage.

jaraco:
Those lines were added in 5daaf7cd967da08a031cb40e3b668a173b4673a1, apparently for [bpo-668662](https://bugs.python.org/issue668662).

hodgestar:
On my Ubuntu 22.04.1 LTS system, this deprecation warning behaves in a very odd way when using Cython's pyximport on Python 3.10.5.

If I create a small test python file, `pyxtest.py`:
```
import pyximport
pyximport.install()
import foo  # noqa
foo.hello()
```
and a small `foo.pyx` file next to it:
```
#cython: language_level=3

def hello():
    print("Hello!")
```
then running the script produces:
```
$ rm -rf ~/.pyxbld ; python -W default::DeprecationWarning: pyxtest.py
/home/user/miniconda3/envs/qutip/lib/python3.10/site-packages/pyximport/pyximport.py:51: DeprecationWarning: the imp module is deprecated in favour of importlib and slated for removal in Python 3.12; see the module's documentation for alternative uses
  import imp
[...]

PR #302, opened 2024-10-08
Fix issue with absolute path with Python 3.13 on Windows

adang1345:
Fix https://github.com/pypa/setuptools/issues/4669.

I originally opened a pull request at https://github.com/pypa/setuptools/pull/4671 but was asked to submit the change here instead. See the above links for details about the bug and the fix.


lazka:
lgtm in general, but the commit message could explain why it broke and why this fixes it.

For future reference, this function tries to implement `os.path.splitroot()[-1]`, but that got only added in 3.12.

adang1345:
I edited the commit message to explain the issue and the fix.

lazka:
better, thanks

FFY00 (approved):

jaraco:
I think this change introduces some lint, so that'll need to be fixed after the other emergent lint issues in main are resolved.

I've left a comment in the original issue - I'd first like to understand if this change to Python 3.13 is indeed intentional. If it is, I'm hoping there's some advice on how to handle this situation, and I'd like to see a documented function that implements this logic in a re-usable way.

distutils/ccompiler.py
```diff
@@ -989,7 +989,8 @@ def _make_relative(base):
         # Chop off the drive
         no_drive = os.path.splitdrive(base)[1]
         # If abs, chop off leading /
-        return no_drive[os.path.isabs(no_drive) :]
+        is_abs = os.path.isabs(no_drive) or sys.platform == 'win32' and (no_drive.startswith('/') or no_drive.startswith('\\'))
```
Avasam, line 992: This can be simplified slightly:
```suggestion
        is_abs = os.path.isabs(no_drive) or sys.platform == 'win32' and no_drive.startswith(('/', "\\"))
```

jaraco:
I started working on adding a regression test for the missed behavior, but it's tricky because in #169 `CCompiler._make_relative` is disabled for tests, so adding a test to `CCompiler._make_out_path` won't exercise the `make-relative` behavior. So maybe we'll do without a test.

jaraco:
> For future reference, this function tries to implement `os.path.splitroot()[-1]`, but that got only added in 3.12.

I've found that `pathlib.Path.relative_to(pathlib.Path.anchor)` works for older Pythons and is consistent across platforms.

PR #307, opened 2024-10-25
type `Distribution.get_command_obj` to not return `None` with `create=True`

Avasam:
This mirrors the typeshed change https://github.com/python/typeshed/pull/11950

This is only visible for distutils itself and setuptools in https://github.com/pypa/setuptools/pull/4704

### DIFF
--- a/distutils/dist.py
+++ b/distutils/dist.py
@@ -15,7 +15,7 @@ import sys
 import warnings
 from collections.abc import Iterable
 from email import message_from_file
-from typing import TYPE_CHECKING, TypeVar, overload
+from typing import TYPE_CHECKING, Literal, TypeVar, overload
 
 from packaging.utils import canonicalize_name, canonicalize_version
 
@@ -31,6 +31,7 @@ from .fancy_getopt import FancyGetopt, translate_longopt
 from .util import check_environ, rfc822_escape, strtobool
 
 if TYPE_CHECKING:
+    # type-only import because of mutual dependence between these modules
     from .cmd import Command
 
 _CommandT = TypeVar("_CommandT", bound="Command")
@@ -837,7 +838,15 @@ Common commands: (see '--help-commands' for more)
 
         raise DistutilsModuleError(f"invalid command '{command}'")
 
-    def get_command_obj(self, command, create=True):
+    @overload
+    def get_command_obj(
+        self, command: str, create: Literal[True] = True
+    ) -> Command: ...
+    @overload
+    def get_command_obj(
+        self, command: str, create: Literal[False]
+    ) -> Command | None: ...
+    def get_command_obj(self, command: str, create: bool = True) -> Command | None:
         """Return the command object for 'command'.  Normally this object
         is cached on a previous call to 'get_command_obj()'; if no command
         object for 'command' is in the cache, then we either create and


distutils/dist.py
```diff
@@ -829,7 +835,11 @@ def get_command_class(self, command):
 
         raise DistutilsModuleError(f"invalid command '{command}'")
 
-    def get_command_obj(self, command, create=True):
+    @overload
+    def get_command_obj(self, command: str, create: Literal[True] = True) -> Command: ...
+    @overload
+    def get_command_obj(self, command: str, create: Literal[False]) -> Command | None: ...
+    def get_command_obj(self, command: str, create: bool = True) -> Command | None:
```
jaraco, line 842: This complicated signature, caused by a boolean parameter, should probably be re-written as two methods, `get_command_obj` and `get_cached_command_obj`, where the latter just returns `self.command_obj.get(command)`. Of course, that's not a concern for this PR.

distutils/dist.py
```diff
@@ -829,7 +835,11 @@ def get_command_class(self, command):
 
         raise DistutilsModuleError(f"invalid command '{command}'")
 
-    def get_command_obj(self, command, create=True):
+    @overload
+    def get_command_obj(self, command: str, create: Literal[True] = True) -> Command: ...
+    @overload
+    def get_command_obj(self, command: str, create: Literal[False]) -> Command | None: ...
```
jaraco, line 841: Is this overload needed? I don't see how it adds any useful information.

Avasam: Without overloads, false-positives (before):
```python
distribution: Distribution = ...
some_command = distribution.get_command_obj("some_command")
# Mypy: Revealed type is "distutils.cmd.Command | None"
# Pyright: Type of "some_command" is "Command | None"
reveal_type(some_command)
# Mypy: Item "None" of "Command | None" has no attribute "some_command_create_true" (union-attr)
# Pyright: "some_command_create_true" is not a known attribute of "None" (reportOptionalMemberAccess)
some_command_create_true.some_command_create_true
```

With overloads (this PR):
```python
distribution: Distribution = ...
some_command = distribution.get_command_obj("some_command")
# Mypy: Revealed type is "distutils.cmd.Command"
# Pyright: Type of "some_command" is "Command"
reveal_type(some_command)
some_command.sub_commands  # OK
```

In both cases, `create=True` doesn't change
```
# True-positive
some_command = distribution.get_command_obj("some_command", False)
# Mypy: Revealed type is "distutils.cmd.Command | None"
# Pyright: Type of "some_command" is "Command | None"
reveal_type(some_command)
# Mypy: Item "None" of "Command | None" has no attribute "sub_commands  " (union-attr)
# Pyright: "sub_commands " is not a known attribute of "None" (reportOptionalMemberAccess)
some_command.sub_commands  
```

jaraco: I don't think that answered my question. So I fumbled around until I was able to somewhat replicate your examples above. I created foo.py:

```python
import distutils.dist

distribution: distutils.dist.Distribution = ...
some_command = distribution.get_command_obj("some_command", False)
# Mypy: Revealed type is "distutils.cmd.Command"
# Pyright: Type of "some_command" is "Command"
reveal_type(some_command)
some_command.sub_commands  # OK
```

And I ran `mypy foo.py` with this copy of distutils, and indeed I got:

```
foo.py:7: note: Revealed type is "Union[distutils.cmd.Command, None]"
```

Then I removed this overload and ran it again, and got a different result:

```
foo.py:4: error: Argument 2 to "get_command_obj" of "Distribution" has incompatible type "Literal[False]"; expected "Literal[True]"  [arg-type]
foo.py:7: note: Revealed type is "distutils.cmd.Command"
```

That's interesting. I would have expected the third definition, the one with `create: bool = True` to fulfill the `Literal[False]` case, instead of having to overload it with exactly the same signature. Alas, it seems our type system isn't so elegant.

Any idea if that's an issue being tracked by the type checking ecosystem?

Avasam: >  I would have expected the third definition

There's only two overload definitions.
The "third" is the implementation. That is (and should) be completely ignored by type-checkers when using that method. Annotating the implementation is only useful to type-check the content of the function.

PR #308, opened 2024-10-28
Make reinitialize_command's return type Generic when "command" argument is a Command

Avasam:
This is a part of adding typeshed's typevars to setuptools, but this change is better handled on distutils' side rather than adding typing-only overloaded overrides in setuptools.

This won't have any effect outside this repo and https://github.com/pypa/setuptools/pull/4704 until https://github.com/pypa/setuptools/issues/4689 , but it also shouldn't have any negative effect by itself.

### DIFF
--- a/distutils/cmd.py
+++ b/distutils/cmd.py
@@ -4,14 +4,19 @@ Provides the Command class, the base class for the command classes
 in the distutils.command package.
 """
 
+from __future__ import annotations
+
 import logging
 import os
 import re
 import sys
+from typing import TypeVar, overload
 
 from . import _modified, archive_util, dir_util, file_util, util
 from ._log import log
 from .errors import DistutilsOptionError
+
+_CommandT = TypeVar("_CommandT", bound="Command")
 
 
 class Command:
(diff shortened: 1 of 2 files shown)


jaraco:
Can you explain why these overloads are necessary? In particular, why is the Generic necessary? What is the difference between accepting a `Command` and a `CommandT` (bound to `Command`)?

Avasam:
> Can you explain why these overloads are necessary? In particular, why is the Generic necessary? What is the difference between accepting a `Command` and a `CommandT` (bound to `Command`)?

With generic:
```python
command: Command = ...
install_lib_command: install_lib = ...
reinitialized_command = command.reinitialize_command(install_lib_command)
# Mypy: Revealed type is "distutils.command.install_lib.install_lib"
# Pyright: Type of "reinitialized_command" is "install_lib"
reveal_type(reinitialized_command)
```

Without generic:
```python
command: Command = ...
install_lib_command: install_lib = ...
reinitialized_command = command.reinitialize_command(install_lib_command)
# Mypy: Revealed type is "distutils.cmd.Command"
# Pyright: Type of "reinitialized_command" is "Command"
reveal_type(reinitialized_command)
```

The generic allows consumers of this method to not loose on type specificity. Otherwise you'd have to `typing.cast` back to the original type to use its subclass-specific features. (as can be seen in setuptools)

PR #310, opened 2024-11-01
Set `Py_GIL_DISABLED=1` for free threaded Python on Windows

colesbury:
When free threaded CPython is installed from the official Windows installer it doesn't have the macro `Py_GIL_DISABLED` properly set becuase its `pyconfig.h` file is shared across the co-installed default build.

Define the macro when building free threaded Python extensions on Windows so that each individual C API extension doesn't have to work around this limitation.

See https://github.com/pypa/setuptools/issues/4662

### DIFF
--- a/distutils/command/build_ext.py
+++ b/distutils/command/build_ext.py
@@ -23,7 +23,7 @@ from ..errors import (
 )
 from ..extension import Extension
 from ..sysconfig import customize_compiler, get_config_h_filename, get_python_version
-from ..util import get_platform, is_mingw
+from ..util import get_platform, is_mingw, is_freethreaded
 
 # An extension name is just a dot-separated list of Python NAMEs (ie.
 # the same as a fully-qualified module name).
@@ -332,6 +332,12 @@ class build_ext(Command):
         # late initialization of compiler even if they shouldn't...)
         if os.name == 'nt' and self.plat_name != get_platform():
             self.compiler.initialize(self.plat_name)
+
+        # The official Windows free threaded Python installer doesn't set
+        # Py_GIL_DISABLED because its pyconfig.h is shared with the
+        # default build, so we need to define it here.
+        if os.name == 'nt' and is_freethreaded():
+            self.compiler.define_macro('Py_GIL_DISABLED', '1')
 
         # And make sure that any compile/link-related options (which might
         # come from the command-line or from the setup script) are set in
--- a/distutils/command/build_ext.py
+++ b/distutils/command/build_ext.py
@@ -335,7 +335,8 @@ class build_ext(Command):
 
         # The official Windows free threaded Python installer doesn't set
         # Py_GIL_DISABLED because its pyconfig.h is shared with the
-        # default build, so we need to define it here.
+        # default build, so we need to define it here
+        # (see pypa/setuptools#4662).
         if os.name == 'nt' and is_freethreaded():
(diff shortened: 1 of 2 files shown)


distutils/command/build_ext.py
```diff
@@ -333,6 +333,12 @@ def run(self):  # noqa: C901
         if os.name == 'nt' and self.plat_name != get_platform():
             self.compiler.initialize(self.plat_name)
 
+        # The official Windows free threaded Python installer doesn't set
+        # Py_GIL_DISABLED because its pyconfig.h is shared with the
+        # default build, so we need to define it here.
```
Zheaoli, line 338: nits: I think it would be better if we put the related link to here

https://github.com/python/cpython/issues/126267
https://github.com/pypa/setuptools/issues/4662

Zheaoli (approved):
LGTM

Zheaoli:
BTW I think we should release a new version that includes this PR ASAP. I think this issue have already been a block reason for the people who want to build free threading extension on windows

rgommers (approved):
In case it helps: this change looks correct to me, and is the right (and probably the only reasonable) choice to make in my opinion. We made the same change in Meson (xref https://github.com/mesonbuild/meson/pull/13338) and have tested it pretty heavily by now - it's working as advertised, while without the `Py_GIL_DISABLED` define, users are getting hard to understand errors at build time.

So +1 for merging this.

vstinner (approved):
LGTM. It's unfortunate that we have to do that, but the fix looks good to me.

FFY00 (approved):
I think the workaround is okay, but we should look into fixing the source of the issue.

FFY00:
The test failures seem unrelated. Unfortunately, they are also failing on `main`, but they don't look the same. I'll rebase this PR on `main` and manually compare the PR tests with the ones from `main`.

FFY00:
The tests seem similar enough, I'll merge this and then see if I have time to fix the CI :upside_down_face:

FFY00:
This specific issue is a symptom of Python not being designed to have paths shared between multiple builds. I think it makes sense to at least discuss in more depth how this could be avoided.

zooba:
> I think it makes sense to at least discuss in more depth how this could be avoided.

Supporting the same set of headers for all platforms (or at minimum, all Windows architectures) and relying on preprocessor switches to choose the build options is a pretty standard way of doing this on Windows.

I certainly wouldn't want to switch wholesale to a configure/make model for all Windows users just for the sake of a temporary flag. If there's a second feature that deserves this, we can consider it.

lysnikolaou:
Hey @FFY00! Are you going to merge this in pypa/setuptools as well? Having a released version of setuptools with this included would help a lot in unblocking CI for a lot of projects.

FFY00:
Hey, sorry for the late reply, I am not a setuptools maintainer.

cc @jaraco

jaraco:
My new job has been keeping me busy, so I'm just getting caught up on this. Let's see...

For future reference, I would have done the merge differently:

- Merges in distutils need to have project-absolute references (e.g. `pypa/distutils#310` instead of `#310`) because these commits are merged into Setuptools and when they're merged, they'll become Setuptools references.
- Also, to help maintain fidelity, this project prefers to use merge commits instead of squash commits, as it provides the ability to summarize changes by looking at merge commits or dive into the details.

I'll get those things repaired and then get the latest changes merged into Setuptools.

PR #311, opened 2024-11-06
Accept any non-string iterable for `distutils.Extension`'s sources

agriyakhetarpal:
## Description

This PR continues what was discussed in https://github.com/python/typeshed/pull/12958#issuecomment-2460661585 on allowing a loosened check for the source files for a `distutils` `Extension` instance.

Previously: just lists were accepted; now: any non-string iterable is accepted (tuples, sets, other iterators, etc.). I plan to update python/typeshed in accordance with this change.

## Additional context

xref: https://github.com/python/mypy/issues/18107, https://github.com/python/typeshed/pull/12958

### DIFF
--- a/distutils/extension.py
+++ b/distutils/extension.py
@@ -115,7 +115,7 @@ class Extension:
                 "'sources' must be an iterable of strings or PathLike objects, not a string"
             )
 
-        # mow we check if it's iterable and contains valid types
+        # now we check if it's iterable and contains valid types
         try:
             self.sources = list(map(os.fspath, sources))
         except TypeError:
--- a/distutils/extension.py
+++ b/distutils/extension.py
@@ -26,12 +26,14 @@ class Extension:
       name : string
         the full name of the extension, including any packages -- ie.
         *not* a filename or pathname, but Python dotted name
-      sources : [string | os.PathLike]
-        list of source filenames, relative to the distribution root
-        (where the setup script lives), in Unix form (slash-separated)
-        for portability.  Source files may be C, C++, SWIG (.i),
-        platform-specific resource files, or whatever else is recognized
-        by the "build_ext" command as source for a Python extension.
+      sources : Iterable[string | os.PathLike]
+        iterable of source filenames (except strings, which could be misinterpreted
+        as a single filename), relative to the distribution root (where the setup
+        script lives), in Unix form (slash-separated) for portability. Can be any
+        non-string iterable (list, tuple, set, etc.) containing strings or
+        PathLike objects. Source files may be C, C++, SWIG (.i), platform-specific
+        resource files, or whatever else is recognized by the "build_ext" command
+        as source for a Python extension.
       include_dirs : [string]
         list of directories to search for C/C++ header files (in Unix
         form for portability)
@@ -106,12 +108,23 @@ class Extension:
     ):
         if not isinstance(name, str):
             raise AssertionError("'name' must be a string")  # noqa: TRY004
-        if not (
-            isinstance(sources, list)
-            and all(isinstance(v, (str, os.PathLike)) for v in sources)
-        ):
+
+        # we handle the string case first; though strings are iterable, we disallow them
+        if isinstance(sources, str):
+            raise AssertionError(  # noqa: TRY004
+                "'sources' must be an iterable of strings or PathLike objects, not a string"
+            )
+
+        # mow we check if it's iterable and contains valid types
+        try:
+            sources = list(sources)  # convert to list for consistency
+            if not all(isinstance(v, (str, os.PathLike)) for v in sources):
+                raise AssertionError(
+                    "All elements in 'sources' must be strings or PathLike objects"
+                )
+        except TypeError:
             raise AssertionError(
-                "'sources' must be a list of strings or PathLike objects."
+                "'sources' must be an iterable of strings or PathLike objects"
             )
 
         self.name = name
--- a/distutils/extension.py
+++ b/distutils/extension.py
@@ -117,18 +117,13 @@ class Extension:
 
         # mow we check if it's iterable and contains valid types
         try:
-            sources = list(sources)  # convert to list for consistency
-            if not all(isinstance(v, (str, os.PathLike)) for v in sources):
-                raise AssertionError(
-                    "All elements in 'sources' must be strings or PathLike objects"
-                )
+            self.sources = list(map(os.fspath, sources))
         except TypeError:
             raise AssertionError(
                 "'sources' must be an iterable of strings or PathLike objects"
             )
 
         self.name = name
-        self.sources = list(map(os.fspath, sources))
         self.include_dirs = include_dirs or []
         self.define_macros = define_macros or []
         self.undef_macros = undef_macros or []
--- a/distutils/extension.py
+++ b/distutils/extension.py
@@ -109,7 +109,7 @@ class Extension:
         if not isinstance(name, str):
             raise AssertionError("'name' must be a string")  # noqa: TRY004
 
-        # we handle the string case first; though strings are iterable, we disallow them
+        # handle the string case first; since strings are iterable, disallow them
         if isinstance(sources, str):
             raise AssertionError(  # noqa: TRY004
                 "'sources' must be an iterable of strings or PathLike objects, not a string"
--- a/distutils/extension.py
+++ b/distutils/extension.py
@@ -107,11 +107,11 @@ class Extension:
         **kw,  # To catch unknown keywords
     ):
         if not isinstance(name, str):
-            raise AssertionError("'name' must be a string")  # noqa: TRY004
+            raise TypeError("'name' must be a string")  # noqa: TRY004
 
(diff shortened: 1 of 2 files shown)


agriyakhetarpal:
I have a bunch of failing tests, linter errors, and several warnings from VS Code extensions locally; I hope the CI will be cleaner in its output. :)

.gitignore
```diff
@@ -0,0 +1,4 @@
+*.egg-info/
+*.pyc
+.coverage
+
```
brianschubert, line 4: Including a `.gitignore` with this PR might not be desired based on https://github.com/pypa/distutils/pull/45#issuecomment-873626642

agriyakhetarpal: Thanks for the context! Removed the file; I'm fine with a global git-ignore configuration. Apologies for the force push.

distutils/extension.py
```diff
@@ -106,12 +108,23 @@ def __init__(
...
+
+        # mow we check if it's iterable and contains valid types
+        try:
+            sources = list(sources)  # convert to list for consistency
+            if not all(isinstance(v, (str, os.PathLike)) for v in sources):
+                raise AssertionError(
+                    "All elements in 'sources' must be strings or PathLike objects"
+                )
+        except TypeError:
             raise AssertionError(
```
Avasam, line 126: Question for the distutils maintainers: I can't help but feel that `AssertionError` is misused here, shouldn't these all be `TypeError` ? (it's noqa'd but not explained)

agriyakhetarpal: I have the same suggestion! :)

brianschubert: Peeking at the history, these used to be actual `assert` statements (https://github.com/pypa/distutils/commit/7d70ca7551824bdf4950932bd5d68e8a982cf1d0), so my guess is that the `AssertionError`s were kept for backwards compatibility with code that might try to catch these errors.

agriyakhetarpal: Additionally, what I infer from the commit message is that that they were added because in-line assert statements are completely removed from the bytecode if used with `-O`

distutils/extension.py
```diff
@@ -106,12 +108,23 @@ def __init__(
...
+                "'sources' must be an iterable of strings or PathLike objects, not a string"
+            )
+
+        # mow we check if it's iterable and contains valid types
+        try:
+            sources = list(sources)  # convert to list for consistency
+            if not all(isinstance(v, (str, os.PathLike)) for v in sources):
+                raise AssertionError(
+                    "All elements in 'sources' must be strings or PathLike objects"
+                )
```
Avasam, line 124: On one hand this prevents issues with single-use iterators, one the other it seems like a lot of iterations over the same source over and over.

`os.fspath` already does type-checking, this entire check just seems useless to me

```suggestion
            self.sources = list(map(os.fspath, sources))
```
(and remove the `self.sources = list(map(os.fspath, sources))` below)

agriyakhetarpal: > `os.fspath` already does type-checking

Interesting, didn't know that! Addressed in 4a467fac7a98cc5d605afc1ebd951cdd82268f86.

distutils/extension.py
```diff
@@ -106,16 +108,22 @@ def __init__(
...
-            and all(isinstance(v, (str, os.PathLike)) for v in sources)
-        ):
+
+        # we handle the string case first; though strings are iterable, we disallow them
+        if isinstance(sources, str):
+            raise AssertionError(  # noqa: TRY004
+                "'sources' must be an iterable of strings or PathLike objects, not a string"
+            )
+
+        # mow we check if it's iterable and contains valid types
```
Avasam, line 118:
```suggestion
        # now we check if it's iterable and contains valid types
```

agriyakhetarpal: Oops, good catch! 😄

Avasam:
To expand on the additional context: This allows typing the `sources` parameter in a way that avoids type invariance issues.
This indirectly relates to https://github.com/pypa/setuptools/issues/2345 and https://github.com/pypa/setuptools/issues/4689

distutils/extension.py
jaraco: I suspect any Exception here is fine, so a TypeError should be fine (famous last words).

distutils/extension.py
```diff
@@ -106,16 +108,22 @@ def __init__(
     ):
         if not isinstance(name, str):
             raise AssertionError("'name' must be a string")  # noqa: TRY004
-        if not (
-            isinstance(sources, list)
-            and all(isinstance(v, (str, os.PathLike)) for v in sources)
-        ):
+
+        # we handle the string case first; though strings are iterable, we disallow them
```
jaraco, line 112: nitpick: Try to use imperative voice in comments.

```suggestion
        # handle the string case first; since strings are iterable, disallow them
```

distutils/extension.py
```diff
@@ -106,16 +108,22 @@ def __init__(
...
+        if isinstance(sources, str):
+            raise AssertionError(  # noqa: TRY004
+                "'sources' must be an iterable of strings or PathLike objects, not a string"
+            )
```
jaraco, line 116: Is it possible instead to leverage the Type Annotations to perform the type validation at run time? Feels redundant to be performing these checks manually. e.g. `@typeguard.typechecked` on `__init__` and add an appropriate annotation for it (and name also)?

Avasam: I it possible? technically yes. The amount of support may also depend on the Python version for introspection of annotations on a function, forward references of annotations, stringified annotations, etc.

This is what a library like [pydantic](https://pypi.org/project/pydantic/) is dedicated to do, as an example.

distutils/tests/test_extension.py
```diff
@@ -69,7 +69,7 @@ def test_extension_init(self):
         assert ext.name == 'name'
 
         # the second argument, which is the list of files, must
-        # be a list of strings or PathLike objects
+        # be a list of strings or PathLike objects, and not a string
```
Avasam, line 72:
```suggestion
        # be an iterable of strings or PathLike objects, and not a string
```

distutils/extension.py
agriyakhetarpal: Thanks!

distutils/extension.py
agriyakhetarpal: For uniformity, I changed the `AssertionError` for both the name and the sources checks in efeb97c02684965d63e78eb9319458b0e8074f66 to be a `TypeError`.

distutils/extension.py
agriyakhetarpal: Sure, I could add an annotated check, though `typeguard` isn't part of the stdlib – is it? 

To add something like `TypeGuard[Iterable]` with a custom validator that wraps this logic, that also means adding a dependency on `typing_extensions` for Python <=3.10, IIRC,

Avasam: > To add something like `TypeGuard[Iterable]` with a custom validator that wraps this logic, that also means adding a dependency on `typing_extensions` for Python <=3.10, IIRC.

`typing.TypeGuard` and `typing.TypeIs` only make sense as return annotations, and I don't think would be used as part of any runtime validation. So as long as you use `from __future__ import annotations` (or quote the annotation), and lock the `typing_extensions` import behind a `TYPE_CHECKING` block, then you can use `typing_extensions.TypeGuard` on any non-EOL python version.

As for adding a dependency on the library `typeguard`, it would indeed also add a dependency on `typing_extensions` (and `importlib_metadata` for Python 3.9)

agriyakhetarpal: Thanks for the correction, yes, my idea was to separate the runtime validation from the development-time type hints but you explained it more concretely than I did. :)

I'll wait for further directions, leaving it to @jaraco and you to suggest if opting for `typeguard` as an optional dependency guarded by `TYPE_CHECKING` would be the best way for `distutils` going forward. 

If there's something specific besides this that I can do within this PR, please let me know!

jaraco: I'd like to use typeguard here. I'm okay with adding dependencies, especially when that means avoiding repetitive type-checking logic that's been automated elsewhere. However, it doesn't need to happen for this PR. I'll track that effort separately. Thanks for the background.

distutils/extension.py
```diff
@@ -26,12 +26,14 @@ class Extension:
...
-        platform-specific resource files, or whatever else is recognized
-        by the "build_ext" command as source for a Python extension.
+      sources : Iterable[string | os.PathLike]
+        iterable of source filenames (except strings, which could be misinterpreted
+        as a single filename), relative to the distribution root (where the setup
+        script lives), in Unix form (slash-separated) for portability. Can be any
+        non-string iterable (list, tuple, set, etc.) containing strings or
+        PathLike objects. Source files may be C, C++, SWIG (.i), platform-specific
+        resource files, or whatever else is recognized by the "build_ext" command
+        as source for a Python extension.
```
Avasam, line 36: I just realized a possible inaccuracy: This documents instance attributes, not parameters. In which case the old docstring was (almost) correct. (sources is still a `list[str]`, not a `list[[string | os.PathLike]` because of `os.fspath`)

PR #315, opened 2024-11-24
Remove py38 compat modules

Avasam:
This also fixes leftover `UP036` from https://github.com/pypa/distutils/pull/314


ruff.toml
```diff
@@ -47,6 +47,10 @@ ignore = [
 	"TRY400",
 ]
 
+[lint.isort]
+combine-as-imports = true
+split-on-trailing-comma = false
+
```
jaraco, line 53: Should these changes go into the skeleton? Why are they unique to this project?

jaraco:
Should we remove the TODO exclusions in ruff.toml?

jaraco:
> Should we remove the TODO exclusions in ruff.toml?

Oh, right, those exclusions weren't present where this change was based. I'll just remove them after merging.

ruff.toml
jaraco: In fc15d4575, I've reverted this change and applied the standard formatting. I'd rather not tussle with ruff's preferences unless there's a strong reason to do so.

PR #318, opened 2024-11-25
Type `Command.sub_commands` to avoid variance issues

Avasam:
This mirrors https://github.com/python/typeshed/pull/11951 and solves an issue revealed by https://github.com/pypa/setuptools/pull/4704


Avasam:
Included in https://github.com/pypa/distutils/pull/319

PR #320, opened 2024-11-25
Coerce Distribution.script_args to list

Avasam:
This will allow any iterable to be passed as `script_args` for `setup`.
After this is merged, typeshed can be updated to reflect this change, which will allow https://github.com/jaraco/jaraco.office/pull/2/files#diff-bb91dce3f1ce94166001a97817fed8d6a6a56c6d7f95233cc28268aea7f05778R24 to pass.

If you don't want the coercion, then `script_args` should at least be typed as `Sequence[str]` and https://github.com/python/typeshed/pull/13116 should be merged first.

I also noticed, I think there's no validation that `script_args` isn't a plain `str`

### DIFF
--- a/distutils/core.py
+++ b/distutils/core.py
@@ -6,9 +6,12 @@ indirectly provides the Distribution and Command classes, although they are
 really defined in distutils.dist and distutils.cmd.
 """
 
+from __future__ import annotations
+
 import os
 import sys
 import tokenize
+from collections.abc import Iterable
 
 from .cmd import Command
 from .debug import DEBUG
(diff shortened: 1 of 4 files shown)


jaraco:
> This will allow any iterable to be passed as `script_args` for `setup`.

Is this the right approach, versus requiring a `list` and requiring consumers to construct lists? I do think `Iterable` is generally preferable, but if this value is a pass-through to get-opt, I'm kinda inclined to just align with what it expects. Nah, this approach seems fine.

> I also noticed, I think there's no validation that `script_args` isn't a plain `str`

I'm fine with that. If the type system allows for "IterableNotStringOrBytesOrDict", that might make sense, but it's kinda gross, so let's just leave it open unless the type system has a solution for that ambiguity.

Avasam:
> if this value is a pass-through to get-opt, I'm kinda inclined to just align with what it expects.

getopt is slightly mistyped atm. It accepts more than just `list`. I need to improve my PR on that, but https://github.com/python/typeshed/pull/13116

Issue #299, opened 2024-09-21
CXXFLAGS overrides default flags, CFLAGS doesn't


mglisse:
https://github.com/pypa/distutils/blob/378984e02edae91d5f49425da8436f8dd9152b8a/distutils/sysconfig.py#L343-L345

distutils now uses CXXFLAGS instead of CFLAGS for C++, that's good. However, it was implemented differently: while CFLAGS is **appended** to the default flags, CXXFLAGS **replaces** them. Is there a particular reason for this inconsistency? I have an application where I was setting both CFLAGS and CXXFLAGS to the same value, and was quite confused when the behavior changed with the new setuptools.

thesamesam:
Thanks @cjwatson. Let me take a look.

cjwatson:
On second thoughts, based on pypa/setuptools#4836, I think this is just a bug in `persistent` that's been exposed by this change.  I've proposed https://github.com/zopefoundation/persistent/pull/214 to fix it.

thesamesam:
Many thanks! That looks right to me and that the assertions were broken (and clearly haven't worked for at least some time).

PR #322, opened 2024-12-22
Use CFLAGS if set as-is, match CXXFLAGS behavior

thesamesam:
Since 2c937116cc0dcd9b26b6070e89a3dc5dcbedc2ae, CXXFLAGS is used as-is if set in the envionment rather than clobbered by whatever CPython happened to be built with.

Do the same for CFLAGS: use it as-is if set in the environment, don't prepend CPython's saved flags.

Fixes: https://github.com/pypa/distutils/issues/299

### DIFF
--- a/distutils/sysconfig.py
+++ b/distutils/sysconfig.py
@@ -340,6 +340,7 @@ def customize_compiler(compiler):
 
         ldshared = _add_flags(ldshared, 'LD')
         ldcxxshared = _add_flags(ldcxxshared, 'LD')
+        cflags = os.environ.get('CFLAGS', cflags)
         cflags = _add_flags(cflags, 'C')
         ldshared = _add_flags(ldshared, 'C')
         cxxflags = os.environ.get('CXXFLAGS', cxxflags)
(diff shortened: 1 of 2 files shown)


jaraco:
That's weird - I'm using the "new merge experience", and it wasn't running the checks for the PR, even after force-pushing the changes after rebasing to main (I didn't want the other commit in there) and pushing a nitpicky linter fix. I had to go to the actions and manually approve the PR there. I guess the new merge experience still needs some work.

thesamesam:
Thanks @jaraco! (I haven't yet used the new merge experience, sounds like it's a good thing I've left it...)

distutils/sysconfig.py
```diff
@@ -340,6 +340,7 @@ def customize_compiler(compiler):
 
         ldshared = _add_flags(ldshared, 'LD')
         ldcxxshared = _add_flags(ldcxxshared, 'LD')
+        cflags = os.environ.get('CFLAGS', cflags)
         cflags = _add_flags(cflags, 'C')
```
mglisse, line 344: Hi, I am a bit confused why you are keeping the `_add_flags(cflags, 'C')` line, CXX doesn't have it (should it?), and it seems redundant (probably duplicates those flags, although I didn't test)?

thesamesam: No, you're right, I'd left it in because I tried something else, and forgot to clean it up. I'll do that in a followup after testing it.

thesamesam: Thanks, submitted https://github.com/pypa/distutils/pull/325. Sorry for the oversight.

PR #327, opened 2025-02-17
Export `PyInit_pkg` for `pkg.__init__` instead of `PyInit___init__`

abravalheri:
This is a possible solution for https://github.com/pypa/setuptools/issues/4826, using an approach similar to the one suggested in https://github.com/python/cpython/issues/80074#issuecomment-1093812266.

The comment https://github.com/pypa/setuptools/issues/4826#issuecomment-2660168154 seems to indicate this approach would work.

@jaraco, the discussion in https://github.com/python/cpython/issues/80074 suggests the change to be implemented in `pypa/setuptools`, but I feel that it is much cleaner if we do it in `pypa/distutils` (otherwise we would have to copy+paste the entirety of the existing `distutils.command.build_ext:build_ext.get_export_symbols` code).

Given the fact that using `stdlib`'s `distutils` is now deprecated and that this is an edge case, I guess it should be fine...
Please let me know if you would like to do something different.

### DIFF
--- a/distutils/command/build_ext.py
+++ b/distutils/command/build_ext.py
@@ -729,7 +729,7 @@ class build_ext(Command):
         provided, "PyInit_" + module_name.  Only relevant on Windows, where
         the .pyd file (DLL) must export the module "PyInit_" function.
         """
-        name = ext.name.split('.')[-1]
+        name = self._get_module_name_for_symbol(ext)
         try:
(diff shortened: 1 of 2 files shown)


jaraco (approved):
Looks good to me. The failure in cygwin appears to be unrelated; I'll track that in a separate issue.

PR #329, opened 2025-02-24
Merge typeshed's `setuptools._distutils` annotations

Avasam:
Merges the annotations from: https://github.com/python/typeshed/tree/main/stubs/setuptools/setuptools/_distutils

This should allow typeshed to not have to expose and duplicate the private module to add types, instead referencing it directly in https://github.com/python/typeshed/tree/main/stubs/setuptools/distutils

It'll also improve base types for setuptools in https://github.com/pypa/setuptools/issues/4689 / https://github.com/pypa/setuptools/pull/4704

`PathLike` unions are repeated often, I could do like I did in setuptools and create runtime aliases: https://github.com/pypa/setuptools/blob/main/setuptools/_path.py


distutils/cmd.py
```diff
@@ -306,7 +322,9 @@ def set_undefined_options(self, src_cmd, *option_pairs):
             if getattr(self, dst_option) is None:
                 setattr(self, dst_option, getattr(src_cmd_obj, src_option))
 
-    def get_finalized_command(self, command, create=True):
+    # NOTE: Because distutils is private setuptools implementation and we don't need to re-expose all commands here,
+    # we're not overloading each and every command possibility.
```
jaraco, line 326:
```suggestion
    # NOTE: Because distutils is private to Setuptools and not all commands are exposed here,
    # not every possible command is enumerated in the signature.
```

distutils/command/bdist.py
```diff
@@ -26,9 +36,10 @@ def show_formats():
 
 class ListCompat(dict[str, tuple[str, str]]):
     # adapter to allow for Setuptools compatibility in format_commands
-    def append(self, item):
+    @deprecated("format_commands is now a dict. append is deprecated.")
+    def append(self, item: object) -> None:
         warnings.warn(
-            """format_commands is now a dict. append is deprecated.""",
+            "format_commands is now a dict. append is deprecated.",
```
jaraco, line 42: Thanks for introducing me to this technique. I don't love it, because it's duplicative and adds a lot of boilerplate in the import section, but if this is the best practice, let's keep it.

jaraco (approved):

Avasam:
Aouch, that's an ugly commit history, that one could've used a squash ^^" Oh well...

jaraco:
> Aouch, that's an ugly commit history, that one could've used a squash ^^" Oh well...

😬 

I tend not to squash merge except when flagged specifically (because a squash is an irreversible rewrite of history that can mask real issues). I don't mind the history, though I do sometimes squash and/or rewrite the history for a nicer story.

Honestly, I'd not worry too much about it. The merge commit itself reflects the PR and anyone uninterested in the details can just follow the first parent.

PR #331, opened 2025-03-04
Fix failing Ruff test

Avasam:
There's a warning-based test that also fails, idk about that.
It's also on Python 3.10 so I'm pretty confident it doesn't have to do with these changes.


jaraco (approved):

jaraco:
In #333, I'm exploring removing that shim altogether.

PR #332, opened 2025-03-08
In build_scripts, unconditionally set shebang to #!python.

jaraco:
In the past, distutils then Setuptools was a (/the) primary installer for packages. It therefore took on the responsibility of rewriting Python shebangs in scripts.

Later, Setuptools became primarily a build backend and defers to installers (pip, uv, ...) to do the rewriting. Yet, the rewrite logic remained in Setuptools and leads to complications (see astral-sh/uv#2744).

This change removes that rewriting logic, replacing it with a static `#!python` shebang to be rewritten by the installers.

I note that the test suite continues to pass unchanged, so the legacy behavior is not captured in the tests. Furthermore, I don't personally install explicit scripts in any of my packages (as they tend to be non-portable). We'll need to find another way to validate the impact of this change.

Closes pypa/setuptools#4863.

### DIFF
--- a/distutils/command/build_scripts.py
+++ b/distutils/command/build_scripts.py
@@ -75,7 +75,7 @@ class build_scripts(Command):
 
         return outfiles, updated_files
 
-    def _copy_script(self, script, outfiles, updated_files):  # noqa: C901
+    def _copy_script(self, script, outfiles, updated_files):
         shebang_match = None
         script = convert_path(script)
         outfile = os.path.join(self.build_dir, os.path.basename(script))
--- a/distutils/command/build_scripts.py
+++ b/distutils/command/build_scripts.py
@@ -5,7 +5,6 @@ Implements the Distutils 'build_scripts' command."""
 import os
 import re
 import tokenize
-from distutils import sysconfig
 from distutils._log import log
 from stat import ST_MODE
 from typing import ClassVar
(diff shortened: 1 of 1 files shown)


hroncok:
This seems to break the deprecated but still possible way of `setup.py build --executable="/usr/bin/python3 -sP"` + `setup.py install` (no external installers involved).

distutils/command/build_scripts.py
```diff
@@ -106,18 +105,8 @@ def _copy_script(self, script, outfiles, updated_files):  # noqa: C901
...
-                    executable = os.path.join(
-                        sysconfig.get_config_var("BINDIR"),
-                        "python{}{}".format(
-                            sysconfig.get_config_var("VERSION"),
-                            sysconfig.get_config_var("EXE"),
-                        ),
-                    )
                 post_interp = shebang_match.group(1) or ''
-                shebang = "#!" + executable + post_interp + "\n"
+                shebang = f"#!python{post_interp}\n"
```
FFY00, line 109: I believe this is the wrong approach. This code belongs to the `build_scripts` code, which may be invoked via the CLI, and even receives a `executable` argument, which is now explicitly ignored.

The `#!python` functionality is part of [PEP 427](https://peps.python.org/pep-0427), and as such, should be implemented in the wheel building code, not here.

FFY00: It seems `bdist_wheel` already implements this functionality itself, so the `build_editable` issue seems like a bug on the `setuptools.build_meta` backend. I believe fixing it should be as simple as setting `executable` there to `python`, just like `bdist_wheel` does.

https://github.com/pypa/wheel/blob/c8f1f09d1abee526851b4e058331db5489a335c4/src/wheel/_bdist_wheel.py#L382-L384

FFY00:
> My personal view is to prioritize non deprecated workflows. If the transition to non-deprecated workflows is not viable for all users that require such feature, I would welcome viable alternative approaches via PR contributions from the interested parties (which solve simultaneously the problem in [pypa/setuptools#4863](https://github.com/pypa/setuptools/issues/4863) and the one described in [#332 (comment)](https://github.com/pypa/distutils/pull/332#issuecomment-2720773262)).

There's a reason we have a `setuptools.build_meta:__legacy__` and `setuptools.build_meta` backend. I don't think we should be making backward-incompatible changes to the legacy code unless strictly necessary.

In this case, it seems the change was exclusively motivated by a feature of the new backend, and it doesn't seem to me like there's a particular technical requirement for changes to the legacy code, so I can't see the justification for this approach.

PR #333, opened 2025-03-08
Remove Python version gate in _make_strs.

jaraco:


jaraco:
I tried removing the behavior altogether, but it's still needed on Windows in Python 3.13.

The reliance on str appears to be in msvc, so maybe it's possible to perform the conversion there instead.

PR #335, opened 2025-03-09
Remove dry run throughout

jaraco:


Avasam:
From https://github.com/pypa/distutils/pull/362 it seems you should be able to remove `func-returns-value` from `disable_error_code` in `mypy.ini`

Avasam:
Nice ! I'll update my affected PRs after the CI issues are resolved.

Issue #336, opened 2025-03-10
Lost `distutils.ccompiler.compiler_class` in latest release


effigies:
As of this morning, my tests have begun failing with:

```
   /home/runner/work/pybids/pybids/.tox/py39-min/lib/python3.9/site-packages/numpy/distutils/__init__.py:24: in <module>
      from . import ccompiler
  _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
  
      import os
      import re
      import sys
      import shlex
      import time
      import subprocess
      from copy import copy
      from distutils import ccompiler
  >   from distutils.ccompiler import (
          compiler_class, gen_lib_options, get_default_compiler, new_compiler,
          CCompiler
      )
  E   ImportError: cannot import name 'compiler_class' from 'distutils.ccompiler' (/home/runner/work/pybids/pybids/.tox/py39-min/lib/python3.9/site-packages/setuptools/_distutils/ccompiler.py)
```

Note that this import is removed from the latest numpy; this only occurs because I test my declared minimum versions using a `lowest-direct` resolution. This means that if I say I support `numpy >=1.22`, then I get `numpy 1.22`, but if setuptools is an indirect dependency, it is allowed to be the latest. (It's infeasible to use a recursive lowest for long-standing packages, since that will currently find dependencies with no minimum stated version or with sdists that predate `requires-python` and attempt to install python 2 packages.)

If necessary, I can pin an older version of setuptools in my `tox.ini`, but if it's possible to re-add this import, it would save doing this across many repositories.

jaraco:
Closing as duplicate of pypa/setuptools#4876.

PR #337, opened 2025-03-10
Restore public symbols missing after compilers move

Avasam:
Closes https://github.com/pypa/distutils/issues/336 and closes https://github.com/pypa/setuptools/issues/4876

- Adds back `distutils.ccompiler.compiler_class` for backwards compatibility
- Removes importing the `msvc` symbol when `from distutils._msvccompiler import *`

### DIFF
--- a/distutils/_msvccompiler.py
+++ b/distutils/_msvccompiler.py
@@ -1,3 +1,5 @@
 from .compilers.C import msvc
 
+__all__ = ["MSVCCompiler"]
+
 MSVCCompiler = msvc.Compiler
--- a/distutils/ccompiler.py
+++ b/distutils/ccompiler.py
@@ -1,17 +1,18 @@
 from .compilers.C import base
 from .compilers.C.base import (
-    CompileError,
-    LinkError,
+    compiler_class,
     gen_lib_options,
     gen_preprocess_options,
     get_default_compiler,
     new_compiler,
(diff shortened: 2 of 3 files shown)


jaraco:
Thanks for this contrib. I don't see where the need for `PLAT_SPEC_TO_RUNTIME` comes from.

Avasam:
I just restored the names that were public in the modules. If you only wanna add `PLAT_SPEC_TO_RUNTIME` if someone asks for it, I can remove it from this PR.

jaraco:
> I simply restored the names that were public in the modules. If you only wanna add `PLAT_SPEC_TO_RUNTIME` if someone asks for it, I can remove it from this PR.

Okay. Yeah, let's remove it. Thanks!

newsfragments/336.bugfix.rst
```diff
@@ -0,0 +1 @@
+Restore symbols missing after compilers move: `distutils.ccompiler.compiler_class` and `distutils._msvccompiler.PLAT_SPEC_TO_RUNTIME` -- by :user:`Avasam`
```
jaraco, line 1: Let's rename this to 4876.bugfix.rst, since it'll get manifest in Setuptools. Thanks.

Avasam: Right, forgot about that.

PR #338, opened 2025-03-10
Mark distutils as `py.typed`

Avasam:
I thought this wouldn't do anything shipped as a setuptools submodule, but it turns out pyright does understand it as part of its support for submodules for namespace packages. So it's actually useful.

The [spec](https://typing.python.org/en/latest/spec/distributing.html) isn't completely clear here. But for type-checkers that understand it: it's a benefit. And for those that don't, nothing changes.

I think `distutils/compilers` is considered a namespace package, hence the need for the second `py.typed`

Without:
image

With:
image

Relates to:
- https://github.com/python/typeshed/pull/13627
- https://github.com/pypa/setuptools/issues/4689
  - https://github.com/pypa/setuptools/pull/4704
  - https://github.com/pypa/setuptools/pull/4861


Avasam:
@jaraco By all means this shouldn't affect users in any way, but should help me with a WIP PR in setuptools and in typeshed.

PR #341, opened 2025-03-11
Remove/replace runtime dependence on `more-itertools`, `jaraco.functools`, and `jaraco.collections`

Sachaa-Thanasius:
A very small diff that would remove 3/4 of distutil's current dependencies.

Imo, this makes distutils easier to understand by replacing opaque helpers with a) well-known stdlib alternatives, or b) obvious and short idiomatic Python. It also should make vendoring distutils easier (e.g. what setuptools does).

(Re. tests: I ran them locally on Windows via tox, and the results before and after were the same — everything passed besides 1 unrelated cygwin thing.)

(I can make an issue about this first if need be — the diff is so small that I figured this wouldn't hurt, though.)

### DIFF
--- a/distutils/sysconfig.py
+++ b/distutils/sysconfig.py
@@ -19,8 +19,6 @@ import sys
 import sysconfig
 from typing import TYPE_CHECKING, Literal, overload
 
-from jaraco.functools import pass_none
-
 from .ccompiler import CCompiler
 from .compat import py39
 from .errors import DistutilsPlatformError
@@ -77,8 +75,9 @@ def _is_parent(dir_a, dir_b):
 
 if os.name == 'nt':
 
-    @pass_none
     def _fix_pcbuild(d):
+        if d is None:
+            return None
         # In a venv, sys._home will be inside BASE_PREFIX rather than PREFIX.
         prefixes = PREFIX, BASE_PREFIX
         matched = (
@@ -147,12 +146,11 @@ def get_python_inc(plat_specific: bool = False, prefix: str | None = None) -> st
     return getter(resolved_prefix, prefix, plat_specific)
 
 
-@pass_none
 def _extant(path):
     """
     Replace path with None if it doesn't exist.
     """
-    return path if os.path.exists(path) else None
+    return (path if os.path.exists(path) else None) if (path is not None) else None
 
 
 def _get_python_inc_posix(prefix, spec_prefix, plat_specific):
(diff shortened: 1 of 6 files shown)


distutils/sysconfig.py
```diff
@@ -147,12 +146,11 @@ def get_python_inc(plat_specific: bool = False, prefix: str | None = None) -> st
     return getter(resolved_prefix, prefix, plat_specific)
 
 
-@pass_none
 def _extant(path):
     """
     Replace path with None if it doesn't exist.
     """
-    return path if os.path.exists(path) else None
+    return (path if os.path.exists(path) else None) if (path is not None) else None
```
effigies, line 153: Simpler:

```suggestion
    return path if path is not None and os.path.exists(path) else None
```

effigies (commented):
Hope drive-by reviews are welcome. This makes sense to me, and I have one small suggestion.

distutils/sysconfig.py
Sachaa-Thanasius: Much simpler. Cheers.

Avasam:
I know Jason is a strong proponent of DRY, even if it means extra simple dependencies.
Aside from whether the changes will be accepted in the first place, the new code looks good to me, as a non-maintainer.

distutils/sysconfig.py
Sachaa-Thanasius: That `exists()` behavior isn't entirely intuitive to me, and I don't see it mentioned in the `os.path` docs, so I'm not confident enough to switch to it without more confirmation (or peer pressure). Plus, I don't want to overstep in this PR by potentially altering any actual behavior — the `is not` check more explicitly matches the behavior of the `@pass_none` decorator. Hope that makes sense.

Avasam: It totally makes sense. I'd say it was more a curiosity.

jaraco:
Yeah, I'm generally not a fan. The problem with the argument here is that if we apply this concept more generally, it means that every project everywhere has to repeat these same concepts. And then when I propose to add something like `pass_none` or `always_iterable` to the stdlib, they'll argue that no one needs it because no one is using it. Instead, I want to use these constructs where they're useful not only for their intrinsic value but also to demonstrate that usefulness.

I'm happy to take the DictStack replacement, and I'd even appreciate a deprecation of DictStack in favor of ChainMap (for those cases where it's a suitable replacement).

If this had been implemented as a series of commits each addressing the separate concerns, I'd have just cherry-picked that commit.

Can you please re-work with just that chnage?

Sachaa-Thanasius:
Sure, I can split the ChainMap usage out.

And I guess this comes down to opinion, but I find the replacement code more readable and explicit without adding much of anything in the way of verbosity (excluding the overloads, since those are necessary on the decorated functions to type them properly anyway). I'll avoid pushing on that for now, though, and leave that a potential future issue.

Issue #334, opened 2025-03-09
Deprecate "dry run" mode


jaraco:
As I'm working to eliminate the reliance of `distutils.compilers` on `distutils` (as part of https://github.com/pypa/setuptools/issues/3622), I find the "dry run" functionality binds everything together. Spawn has its own behavior to account for dry-run. Each command and compiler has a `.execute` method to honor dry run.

And it got me thinking - does anybody use this functionality? My guess is probably not, especially now that most of this is handled in isolated environments and as part of build backends.

So let's explore just removing it.

jaraco:
I did [a search](https://github.com/search?q=content%3A.dry_run+path%3Asetup.py&type=code), and it does seem there are quite a few references to `--dry-run` being used. Most of them seem like casual use. I couldn't find anything that looked like popular and programmatic use. And since setup.py is deprecated, it should be safe to remove features of setup.py without warning.

Avasam:
Your query link seems to be messed-up and double-encoded, this should be correct: https://github.com/search?q=content%3A%2Fsetup%5C.py.*--dry-run%2F&type=code

As for programatic usage: https://github.com/search?q=content%3A.dry_run+path%3Asetup.py&type=code

Glad to say this is no longer a concern for pywin32 as of https://github.com/mhammond/pywin32/pull/2390/files#diff-60f61ab7a8d1910d86d9fda2261620314edcae5894d5aaa236b821c7256badd7L871 😄

jaraco:
> our query link seems to be messed-up and double-encoded, this should be correct: https://github.com/search?q=content%3A%2Fsetup%5C.py.*--dry-run%2F&type=code

Thanks (fixed). Not sure how that happened.

Avasam:
I think between https://github.com/pypa/distutils/commit/a1c0ebc14a1a080b0b2233634106fb8cbb4c231e and https://github.com/pypa/distutils/commit/a1c0ebc14a1a080b0b2233634106fb8cbb4c231e this is pretty much done ?

PR #343, opened 2025-03-12
WIP: Fix most mypy issues

Avasam:


distutils/sysconfig.py
```diff
@@ -246,14 +246,6 @@ def get_python_lib(
     sys.base_exec_prefix -- i.e., ignore 'plat_specific'.
     """
 
-    if IS_PYPY and sys.version_info < (3, 8):
```
Avasam, line 249: A separate PR to remove old code paths would be a good idea

distutils/tests/test_check.py
```diff
@@ -180,14 +174,7 @@ def foo():
...
             cmd.check_restructuredtext()
             msgs = cmd._check_rst_data(rest_with_code)
-            if pygments is not None:
-                assert len(msgs) == 0
-            else:
-                assert len(msgs) == 1
-                assert (
-                    str(msgs[0][1])
-                    == 'Cannot analyze code. Pygments package not found.'
-                )
```
Avasam, line 190: I don't see pygments being use anywhere or even installed as a transitive dependency. Is that just old code ?

distutils/util.py
```diff
@@ -151,7 +151,9 @@ def change_root(
         if not os.path.isabs(pathname):
             return os.path.join(new_root, pathname)
         else:
-            return os.path.join(new_root, pathname[1:])
+            # type-ignore: This makes absolutes os.Pathlike unsupported in this branch.
+            # Either this or we don't support bytes-based paths, or we complexify this branch.
+            return os.path.join(new_root, pathname[1:])  # type: ignore[index]
```
Avasam, line 156: There's a decision to be taken here.
1. Don't support `PathLike` in this branch (current behaviour)
2. Don't support `bytes | PathLike[bytes]`
```suggestion
            # type-ignore: This makes absolutes os.Pathlike unsupported in this branch.
            # Either this or we don't support bytes-based paths, or we complexify this branch.
            return os.path.join(new_root, str(pathname)[1:])
```
3. Support both, requires extra checking.

distutils/cmd.py
```diff
@@ -330,7 +332,8 @@ def get_finalized_command(self, command: str, create: bool = True) -> Command:
         'command', call its 'ensure_finalized()' method, and return the
         finalized command object.
         """
-        cmd_obj = self.distribution.get_command_obj(command, create)
+        # TODO: Raise a more descriptive error when cmd_obj is None ?
+        cmd_obj = cast(Command, self.distribution.get_command_obj(command, create))
```
Avasam, line 336: Decision to be taken here.

distutils/util.py
Avasam: Extracted to https://github.com/pypa/distutils/pull/363/files#r2072652049

Avasam:
I split up all individual mypy code fixes into their own PRs for ease of review.
The `newer_pairwise` issue mentioned at https://github.com/pypa/distutils/pull/343#issuecomment-2774396602 still needs to be looked at.

distutils/cmd.py
jaraco: My instinct - get_command_obj should raise the error and always return Command... unless there's a case where None is legitimately expected, in which case this should be wrapped in a helper to raise that error and set the expectation.

jaraco: I just did a quick scan through the distutils and setuptools codebase and there's not a single call to `get_finalized_command(..., create=False)`. Maybe the create flag is just cruft that can be eliminated.

distutils/dir_util.py
```diff
@@ -57,10 +57,11 @@ def mkpath(name: pathlib.Path, mode=0o777, verbose=True, dry_run=False) -> None:
...
 
-    try:
-        dry_run or name.mkdir(mode=mode, parents=True, exist_ok=True)
-    except OSError as exc:
-        raise DistutilsFileError(f"could not create '{name}': {exc.args[-1]}")
+    if not dry_run:
+        try:
+            name.mkdir(mode=mode, parents=True, exist_ok=True)
+        except OSError as exc:
+            raise DistutilsFileError(f"could not create '{name}': {exc.args[-1]}")
```
jaraco, line 64: I really don't love how this code gets more nested. What is it about mypy that's forcing different logic here?

distutils/fancy_getopt.py
```diff
@@ -249,6 +249,7 @@ def getopt(self, args: Sequence[str] | None = None, object=None):  # noqa: C901
             raise DistutilsArgError(msg)
 
         for opt, val in opts:
+            value: int | str = val
```
jaraco, line 252: This is gross. `val` and `value` are the same thing. Mypy is forcing some idiosyncratic code. There's got to be a better way (cast `opts.__iter__` to the expected types?).

distutils/sysconfig.py
```diff
@@ -324,6 +324,14 @@ def customize_compiler(compiler: CCompiler) -> None:
...
             'ARFLAGS',
         )
+        assert isinstance(cc, str)
+        assert isinstance(cxx, str)
+        assert isinstance(cflags, str)
+        assert isinstance(ccshared, str)
+        assert isinstance(ldshared, str)
+        assert isinstance(ldcxxshared, str)
+        assert isinstance(ar_flags, str)
+        assert isinstance(shlib_suffix, str)
```
jaraco, line 334: I'm not sure these assertions will always pass. There have been some bugs around these being undefined in unusual environments. Maybe it makes sense to catch these conditions early. Do we have good reason to believe that non-string values returned here would never be viable (would be breaking anyway)?

distutils/tests/unix_compat.py
```diff
@@ -1,10 +1,15 @@
...
+from types import ModuleType
 
+grp: ModuleType | None = None
+pwd: ModuleType | None = None
 try:
     import grp
     import pwd
 except ImportError:
-    grp = pwd = None
+    pass
```
jaraco, line 12: I don't love the asymmetry of this syntax, but if this is the new Pythonic style, I guess it's acceptable. Is it?
It's starting to feel like we need a meta language to express Python imports. It would be so much nicer if we could simply express:
```
opt_import grp
opt_import pwd
```

jaraco, line 12: I'd be tempted to use `contextlib.suppress(ImportError)` to eliminate the no-op block... except that requires a prior import, which may lead to problems. Also, I give it a 50% chance that mypy recognizes that approach.

jaraco (commented):
Leaving an initial review. I've gone over most of this before I realized that portions have been extracted elsewhere.

distutils/cmd.py
Avasam: I opened an exploratory PR deprecating (unused, but not removed, for backward compatibility) the `create` arg: https://github.com/pypa/distutils/pull/377

distutils/dir_util.py
Avasam: This one could be ignored on the line itself without being unsafe.

It's an unused expression, both mypy and pyright warn here because of the lack of assignement, it "looks like" a possible mistake because the result of the expression is not used for anything. But it's just a conditional execution pattern.

It's definitely more of a lint than a type error.

- https://github.com/microsoft/pylance-release/blob/main/docs/diagnostics/reportUnusedExpression.md
- https://github.com/microsoft/pyright/blob/main/docs/configuration.md#user-content-reportunusedexpression
- https://mypy.readthedocs.io/en/latest/error_code_list.html#code-func-returns-value

FWIW, https://github.com/pypa/distutils/issues/334 / https://github.com/pypa/distutils/pull/335 would remove the condition anyway.

distutils/fancy_getopt.py
Avasam: Actually, this may be resolved in mypy 1.18: https://mypy.readthedocs.io/en/stable/changelog.html#flexible-variable-definitions-update

Before 1.18 the flag wasn't stabilized and called `allow-redefinition-new` (I wasn't aware of its existence either)

Edit: No it doesn't help here because of the difference of scope

distutils/sysconfig.py
Avasam: (same comment in the split PR, for tracking: https://github.com/pypa/distutils/pull/366#discussion_r2440251813)

distutils/tests/unix_compat.py
Avasam: There's definitely much to improve when it comes to static type checkers and conditional imports... (especially with `ImportError`/`ModuleNotFoundError`)

Hopefully ty and/or pyrefly will come along with improved semantics and force existing checkers to improve there.

Avasam: ~~It's possible that this can be solved using mypy 1.18's `allow-redefinition`~~
Edit: No it doesn't help here.

distutils/fancy_getopt.py
Avasam: There's this related request https://github.com/python/mypy/issues/15988
Which would allow `val: str | int = val` which is much safer than a cast.

As for casting, I could do
```python
        # Unsafely casting to avoid using a different variable
        # python/mypy#15988 could allow `val: str | int = val`
        for opt, val in cast("list[tuple[str, int | str]]", opts):
```

Avasam:
@jaraco Thanks, I took the time to go through your review even though most things here have been split-up in smaller self-contained PRs. (we can always link back to discussions here)

I think most of your comments/concern are extracted to https://github.com/pypa/distutils/pull/368, so you can probably ignore that one for now.

As for the failing CI, there's a related issue opened at setuptools: https://github.com/pypa/setuptools/issues/5094

distutils/fancy_getopt.py
jaraco: Yes, that's preferable, as it puts the onus on the type system and not the user's code.

Avasam:
Everything split up in smaller PRs, all conversations have been migrated to said PRs, or new issues

PR #344, opened 2025-03-12
Make sure `type(None)` is used instead of `None` for the type argument of `isinstance`.

Sachaa-Thanasius:
Using None, whether alone or in a tuple, as the second argument of isinstance(...) raises a TypeError at runtime. It's not a valid type.

### DIFF
--- a/distutils/compilers/C/base.py
+++ b/distutils/compilers/C/base.py
@@ -237,7 +237,7 @@ class Compiler:
         """
         A valid macro is a ``name : str`` and a ``value : str | None``.
         """
-        return isinstance(name, str) and isinstance(value, (str, None))
+        return isinstance(name, str) and isinstance(value, (str, type(None)))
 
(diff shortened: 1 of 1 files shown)


jaraco:
This regression was added in 19e95750f3465db016708f36f9ee51d155631994 (#329) (cc @Avasam). In the future, we should be careful not to change the behavior when applying type annotations, maybe even do an AST diff to highlight behavior changes hidden in a haystack of annotations.

Avasam:
Thanks for catching the regression.

I'm not sure why that change was even done in my PR, it doesn't affect static typing in any way.

What I think happened is that I used VSCode's PyLance to autofill some return type annotations, it likely autofilled with `types.NoneType`. Then I must've search&replaced `types.NoneType` -> `None` and missed in self-review that it introduced that change.

Worst part is it's even caught by pyright/pylance >.<
image

PR #345, opened 2025-03-15
Explicit re-exports of submodules in `distutils.command.__all__`

Avasam:
This fixes an issue that makes `mypy.stubtest` fail when trying to compare stubs with this runtime.
This is also flagged by pyright's [reportUnsupportedDunderAll](https://github.com/microsoft/pyright/blob/main/docs/configuration.md#reportUnsupportedDunderAll)

Aesthetic change: I sorted the content of `__all__` to match the imports.


abravalheri:
@jaraco, in https://github.com/pypa/setuptools/issues/4902 I hypothesized that eagerly importing the command modules in `command/__init__.py` affected the end result of `setuptools.monkey.patch_all`:

- Previous to this PR the modules were imported late, so by the time they were executed `distutils.core.Command` was already patched to `setuptools.Command`.
- Now, the modules are executed before the patch.

The following experiment seems to corroborate this conclusion:

```bash
> docker run --rm -it python:3.10-bookworm /bin/bash
python -m pip install 'setuptools==77.0.1'
python
>>> import setuptools
>>> from distutils.command.sdist import sdist
>>> sdist.__mro__
(<class 'distutils.command.sdist.sdist'>, <class 'setuptools.Command'>, <class 'distutils.cmd.Command'>, <class 'object'>)


python -m pip install 'setuptools @ https://github.com/pypa/setuptools/archive/refs/tags/v77.0.2.zip'
python
>>> import setuptools
>>> from distutils.command.sdist import sdist
>>> sdist.__mro__
(<class 'distutils.command.sdist.sdist'>, <class 'distutils.cmd.Command'>, <class 'object'>)
```

I don't know if there are the deeper unforeseen consequences of this change (other than problems similar to https://github.com/pypa/setuptools/issues/4902).

Avasam:
@abravalheri Instead of a revert, could you simply put the imports behind a `if TYPE_CHECKING` block for static tooling? (not 100% sure if that's enough for stubtest or not, but at least no more early import at runtime)

jaraco:
I accepted this PR without thinking about it too hard, thinking it was just some syntactic cleanup, but I see now it's meaningfully changed the behavior.

The problem is that mypy and pyright expect a very regimented, uninteresting use of `__all__`. Probably if written today, `__all__` would better be replaced by a distutils-specific property (`all_` maybe).

> @abravalheri Instead of a revert, could you simply put the imports behind a `if TYPE_CHECKING` block for static tooling? (not 100% sure if that's enough for stubtest or not, but at least no more early import at runtime)

Let's confirm that it's enough for stubtest. I don't want to employ a fix that has no effect. My instinct, let's just revert this change for now and work out the appropriate solution in light of the troubles. I still need to review the proposed changes and the bug report, but I'll get something out soon.

PR #346, opened 2025-03-18
Restores _default_compilers to ccompilers.py for backwards compatibility

ManiacDC:
Additional changes for https://github.com/pypa/distutils/issues/336 and https://github.com/pypa/setuptools/issues/4876

Adds back distutils.ccompiler._default_compilers for backwards compatibility

When importing the CC class from numba.pycc, numpy 1.23.5 will try to access distutils.ccompiler._default_compilers directly. This version still supports python 3.10, so it would be preferable not to deprecate this functionality until then.


distutils/ccompiler.py
```diff
@@ -6,18 +6,20 @@
...
-    'show_compilers',
+    "CompileError",
+    "LinkError",
+    "compiler_class",
+    "gen_lib_options",
+    "gen_preprocess_options",
+    "get_default_compiler",
+    "new_compiler",
+    "show_compilers",
+    _default_compilers,
```
jaraco, line 22: This needs to be a string.
```suggestion
    "_default_compilers",
```

jaraco:
> [numpy 1.23.5] still supports python 3.10, so it would be preferable not to deprecate this functionality until then.

I'm not sure I'm following. Setuptools/distutils supports Python 3.9+. Following the bugfix/workaround for numba, what use-cases would benefit from this change? I'd really like to avoid carrying tech debt where it's not adding significant value.

distutils/ccompiler.py
ManiacDC: Yes it does, thanks! I was struggling getting the environment set up correctly and there were so many red and yellow lines I got distracted :(

ManiacDC:
> > [numpy 1.23.5] still supports python 3.10, so it would be preferable not to deprecate this functionality until then.
> 
> I'm not sure I'm following. Setuptools/distutils supports Python 3.9+. Following the bugfix/workaround for numba, what use-cases would benefit from this change? I'd really like to avoid carrying tech debt where it's not adding significant value.

I do understand your stance, I hate tech debt too. Any older packages using numpy.distutils (this is not limited to numba<0.57.0, there are a few hits that can be found via a quick google search) would no longer function without this attribute exported. I do not know how widespread these packages are. numpy.distutils is supported until python 3.11 is dropped.

ManiacDC:
This build failed again, I looked and failed to find how to set up my environment for distutils (packages needed for formatting, linting, and sorting). I think I[ found it now](https://setuptools.pypa.io/en/latest/development/developer-guide.html) (I was looking in distutils, should have been looking in setuptools), I will try to address this tomorrow:

jaraco:
> > > [numpy 1.23.5] still supports python 3.10, so it would be preferable not to deprecate this functionality until then.
> > 
> > 
> > I'm not sure I'm following. Setuptools/distutils supports Python 3.9+. Following the bugfix/workaround for numba, what use-cases would benefit from this change? I'd really like to avoid carrying tech debt where it's not adding significant value.
> 
> I do understand your stance, I hate tech debt too. Any older packages using numpy.distutils (this is not limited to numba<0.57.0, there are a few hits that can be found via a quick google search) would no longer function without this attribute exported. I do not know how widespread these packages are. numpy.distutils is supported until python 3.11 is dropped.

Sounds good. Once it's testing properly, I'll update it a bit to add comments to the imports to explain why they're there (and when they can be removed).

m-gangloff:
I have the exact described issue with the [faiss](https://github.com/facebookresearch/faiss/blob/0f2035cc8369d6475166a564e650fdf9c1cb3299/faiss/python/loader.py#L43) library when importing numpy.distutils.cpuinfo. I'm running on Python 3.10 and numpy 1.26.4 and setuptools 77.0.1.
```shell
  File "<string>", line 1, in <module>
  File "/opt/faiss/build/faiss/python/faiss/__init__.py", line 16, in <module>
    from .loader import *
  File "/opt/faiss/build/faiss/python/faiss/loader.py", line 88, in <module>
    instruction_sets = supported_instruction_sets()
  File "/opt/faiss/build/faiss/python/faiss/loader.py", line 53, in supported_instruction_sets
    if is_sve_supported():
  File "/opt/faiss/build/faiss/python/faiss/loader.py", line 43, in is_sve_supported
    import numpy.distutils.cpuinfo
  File "/usr/local/lib/python3.10/dist-packages/numpy/distutils/__init__.py", line 26, in <module>
    from . import ccompiler
  File "/usr/local/lib/python3.10/dist-packages/numpy/distutils/ccompiler.py", line 733, in <module>
    ccompiler._default_compilers += (('linux.*', 'intel'),
AttributeError: module 'distutils.ccompiler' has no attribute '_default_compilers'. Did you mean: 'get_default_compiler'?
```

ManiacDC:
@jaraco Thank you. I think I have it straightened out now (I couldn't get tox to run fully locally, probably due to improper setup, but the ruff checks pass now).
I also added some comments regarding why these imports were added and when they were removed. Feel free to edit/replace/keep since you said you were planning to add some.

jaraco:
I'm not committing to keeping this compatibility until Python 3.11 is dropped, but we can keep it for some time to alleviate the pressure. Are there versions of numpy that run on older Pythons without numpy.distutils?

ManiacDC:
> I'm not committing to keeping this compatibility until Python 3.11 is dropped, but we can keep it for some time to alleviate the pressure. Are there versions of numpy that run on older Pythons without numpy.distutils?

Thanks for merging this @jaraco! Sorry for the quote changes, I think black automatically switched them and I didn't notice.

numpy.distutils is still in the current version of numpy and it has this issue. I submitted an issue there, but I'm not sure if anything will get done about it until they remove it:
https://github.com/numpy/numpy/issues/28551

PR #347, opened 2025-03-19
Honor but deprecate newlines when processing keywords/platforms

### LINKED ISSUE #4887: [BUG] multiline `keywords` breaks metadata file
### setuptools version

74.1.3 and 76.0.0

### Python version

Python 3.13

### OS

Fedora Linux

_No response_

### Description

I recently discovered that if a project has `keywords` specified as a multiline string, the `METADATA` file generated from it is invalid in the sense that the headers/body split happens on the wrong line.

Example project: https://foss.heptapod.net/python-libs/passlib/-/blob/branch/stable/setup.py?ref_type=heads#L93
But you can find more like this here: https://grep.app/search?f.lang=Python&q=keywords%3D%22%22%22

Metadata generated with setuptools 74.1.3:
```
Provides-Extra: bcrypt
Requires-Dist: bcrypt >=3.1.0 ; extra == 'bcrypt'
Provides-Extra: build_docs
Requires-Dist: sphinx >=1.6 ; extra == 'build_docs'
Requires-Dist: sphinxcontrib-fulltoc >=1.2.0 ; extra == 'build_docs'
Requires-Dist: cloud-sptheme >=1.10.1 ; extra == 'build_docs'
Provides-Extra: totp
Requires-Dist: cryptography ; extra == 'totp'

crypt md5-crypt
sha256-crypt sha512-crypt pbkdf2 argon2 scrypt bcrypt
apache htpasswd htdigest
totp 2fa
Classifier: Intended Audience :: Developers
Classifier: License :: OSI Approved :: BSD License
Classifier: Natural Language :: English
…
```
Here, the empty line ends the headers part, and the key/value pair below it are unreachable in a `Message` instance parsed by `email.parser`.

Metadata generated with setuptools 76.0.0:
```
Name: passlib
Version: 1.7.4
Summary: comprehensive password hashing framework supporting over 30 schemes
Home-page: https://passlib.readthedocs.io
Download-URL: https://pypi.python.org/packages/source/p/passlib/passlib-1.7.4.tar.gz
Author: Eli Collins
Author-email: elic@assurancetechnologies.com
License: BSD
Keywords: password secret hash security
crypt md5-crypt
sha256-crypt sha512-crypt pbkdf2 argon2 scrypt bcrypt
apache htpasswd htdigest
totp 2fa
Classifier: Intended Audience :: Developers
Classifier: License :: OSI Approved :: BSD License
…
```

There is no empty line here but the `keyword` header is not properly folded as a multiline header according to RFC 822 and therefore the result is the same - unreachable key/value pairs below it.

I know there were several discussions about the formatting of different fields so I'm not sure this is something to fix here or not. The documentation says that `keywords` can be:
> comma-separated string providing descriptive meta-data.

so multiline string does not seem to be prohibited.

### Expected behavior

setuptools could either fail with an error message saying that keywords are improperly formatted or automatically replace all newlines with a space.

### How to Reproduce

1. Build a wheel from one of the projects mentioned above.
2. Unpack the wheel.
3. Parse the `METADATA` file manually and check available keys.

### Output

See the above examples.

frenzymadness:
Newlines in `keywords` or `platforms` can break
the produced metadata in PKG-INFO or METADATA files.

Fixes: https://github.com/pypa/setuptools/issues/4887

Another possibility is to raise an error if `keywords` or `platforms` contain newlines but this fix is simple and backward-compatible and does not break projects that have multiline definitions of those attributes.


jaraco:
Thanks for this proposal. My instinct is it would be better to guide users toward one best format, preferably something with proper syntax support, i.e. a full sequence of values, whether in Python or toml. What do you think about instead deprecating anything but `sequence[str]` for keywords & platforms?

distutils/dist.py
```diff
@@ -644,6 +644,7 @@ def finalize_options(self) -> None:
             if value is None:
                 continue
             if isinstance(value, str):
+                value = value.replace("\n", " ")
```
jaraco: This line deserves a comment explaining its purpose (and linking to the downstream setuptools issue). I might even go so far as it should be a function with a docstring describing its purpose, which would have the added benefit of avoiding the need to mutate the `value` variable.

newsfragments/4887.bugfix.rst
```diff
@@ -0,0 +1 @@
+Fixed handling of keywords following the old specification with spaces instead of commas and containing newlines.
```
jaraco: It would be nice to reference the "old specification". Was there once a specification that allowed for whitespace for keyword separators?

jaraco: I'm not sure I'd characterize this change as a bugfix. I don't think it was ever previously intended to support newlines in this field. This change therefore _adds_ support for newlines in keywords (accommodating the incidental use by some packages). Is there any evidence that this form (whitespace separated keywords) was intended to be supported?

distutils/dist.py
jaraco: A wrapper function would have the added benefit that it could be the location for a deprecation warning to remove this support.

jaraco: I see that we're now interpreting `\n` as a space, even though `,` is the proper keyword separator. I'm not sure that honor's the user's intent. Surely if they're supplying keywords on separate lines, they expect them to be separate keywords and not joined to other keywords. Should `\n` imply `,` instead?

newsfragments/4887.bugfix.rst
jaraco: From Claude:

I'll look at that PR to find the changelog reference you mentioned.

Let me dig into what this "old specification" of space-separated keywords actually was. Let me look at the referenced issue and the relevant packaging PEPs.

The two old PEPs both show space-separated keywords. Let me confirm when the spec switched to comma-separated.

I found the full picture. Here's what the "old specification" refers to.

## The "old specification" = space-separated keywords

The changelog entry in [pypa/distutils#347](https://github.com/pypa/distutils/pull/347) says:

> *"Fixed handling of keywords following the old specification with spaces instead of commas and containing newlines."*

The "old specification" is the original Python packaging **core metadata** definition of the `Keywords` field, which specified keywords as **space-separated**, not comma-separated. This traces back through three PEPs:

| Spec | Metadata version | `Keywords` example |
|------|-----------------|--------------------|
| [PEP 241](https://peps.python.org/pep-0241/) (2001) | 1.0 | `Keywords: dog puppy voting election` |
| [PEP 314](https://peps.python.org/pep-0314/) | 1.1 | `Keywords: dog puppy voting election` |
| [PEP 345](https://peps.python.org/pep-0345/) | 1.2 | `Keywords: dog puppy voting election` |

All three showed **space-separated** keywords in their examples.

## The divergence

Despite what the specs said, distutils and setuptools always **implemented** the field as **comma-separated**. Because those tools were so widely used, the de-facto comma behavior won out, and the specification was eventually rewritten to match the tools. The [current core metadata spec](https://packaging.python.org/en/latest/specifications/core-metadata/) now reads:

> *"A list of additional keywords, separated by commas..."* — example: `Keywords: dog,puppy,voting,election`

...and it carries an explicit note that the spec *previously* showed space-separated keywords but was updated to match the comma-based reality of distutils/setuptools.

## Why it mattered for that PR

The PR ([distutils#347](https://github.com/pypa/distutils/pull/347), from setuptools issue [#4887](https://github.com/pypa/setuptools/issues/4887)) fixes a case where `keywords`/`platforms` values contain **newlines**, which corrupt the RFC 822 header structure of `PKG-INFO`/`METADATA`. The changelog's phrase "following the old specification with spaces instead of commas" describes users who still write keywords the old space-separated way (often as a multi-line string), and the fix replaces newlines with spaces (`value.replace("\n", " ")`).

So the "old specification" isn't a single lost document — it's the pre-revision wording of the core metadata standard (PEPs 241 → 314 → 345) that defined `Keywords` as space-separated, before it was reconciled with the comma-separated behavior distutils/setuptools had shipped all along.

One nuance worth flagging (which jaraco raised on the PR): honoring the *old* space-separated intent would arguably mean turning newlines into **commas**, not spaces — since under the current spec each newline-separated line is likely meant to be a distinct keyword.

jaraco: This all makes me think that spaces were _never_ supported by Setuptools, so we shouldn't add support for them now. However, because packages have come to rely on the acceptance of space-separated keywords (and even newlines), we should instead warn the user.

Let's do this:

- Repair newlines, replacing them with commas, and warning the user that it's an invalid input and will error in the future.

I was also tempted to say we should also warn the user if there's exactly one keyword with multiple spaces in it (a heuristic indicating that the user has accidentally used the wrong format for keywords), but now that I think about it, it's probably not worth the trouble.

jaraco: Also, let's hyperlink to PEP 345 for the "old specification".

jaraco: Evidence is that the spec suggested spaces, but spaces were never supported by Setuptools/distutils. Still, let's go with `feature` for this change, and describe the change as "handling improper newline-separated keywords and deprecating that support" (or similar).

pfmoore: The old spec allowed spaces, but not (by my reading, at least) newlines. So I'd consider this as "taking a forgiving approach to encountering invalid input".

distutils/tests/test_dist.py
```diff
@@ -217,6 +217,28 @@ def test_finalize_options(self):
...
+        }
+        with pytest.warns(DeprecationWarning, match="Newlines"):
+            dist = Distribution(attrs=attrs)
+        assert dist.metadata.platforms == ['one', 'two', 'three', 'four']
+        assert dist.metadata.keywords == ['one', 'two', 'three', 'four']
+
+        # Surrounding newlines (e.g. from a triple-quoted string) don't
+        # introduce empty items.
+        attrs = {
+            'keywords': '\none two\nthree four\n',
```
jaraco: What about two newlines? That'll end up creating blank keywords. Let's capture that condition also, and make sure no empty values are generated.

distutils/dist.py
```diff
@@ -74,6 +74,31 @@ def _ensure_list(value: str | Iterable[str], fieldname) -> str | list[str]:
...
+    PEP 345, with spaces). Newlines were never a valid separator and corrupt
+    the generated ``PKG-INFO``/``METADATA`` (pypa/setuptools#4887). Some
+    projects nonetheless supply these fields as newline-separated strings
+    (e.g. a triple-quoted string with one item per line), so treat each line
+    as a separate item by replacing newlines with commas, and warn that the
+    behavior is deprecated.
+    """
+    if "\n" not in value:
+        return value
+    warnings.warn(
```
jaraco, line 98: I'm not sure a proper DeprecationWarning will work for Distutils and/or Setuptools. Deprecation warnings often only emit when running under test. Look at the codebase to determine the best way to warn in this situation so that users of setuptools will see it.

distutils/dist.py
```diff
@@ -644,7 +672,11 @@ def finalize_options(self) -> None:
             if value is None:
                 continue
             if isinstance(value, str):
-                value = [elm.strip() for elm in value.split(',')]
+                value = [
+                    stripped
+                    for elm in _repair_newlines(value).split(',')
+                    if (stripped := elm.strip())
+                ]
```
jaraco: I feel like this code has been written a million times, once for each package that's out there. Let's use `jaraco.text.SeparatedValues` instead.

PR #352, opened 2025-03-21
Only import modules in `distutils/command/__init__.py` if type checking

### LINKED ISSUE #351: Recent changes in `disutils/command/__init__.py` change the outcome of `setuptools.monkey.patch_all`
I am not sure if this new behaviour is welcome, but I am reporting it as an outcome of the investigation in https://github.com/pypa/setuptools/issues/4902.

---

In https://github.com/pypa/setuptools/issues/4902 I hypothesized that eagerly importing the command modules in `command/__init__.py` affected the end result of `setuptools.monkey.patch_all`:

- Previous to this PR the command modules were imported late, so by the time they were executed `distutils.core.Command` was already monkeypatched to `setuptools.Command`.
- Now, the modules are executed before the patch.

The following experiment seems to corroborate this conclusion:

```bash
> docker run --rm -it python:3.10-bookworm /bin/bash
python -m pip install 'setuptools==77.0.1'
python
>>> import setuptools
>>> from distutils.command.sdist import sdist
>>> sdist.__mro__
(<class 'distutils.command.sdist.sdist'>, <class 'setuptools.Command'>, <class 'distutils.cmd.Command'>, <class 'object'>)


python -m pip install 'setuptools @ https://github.com/pypa/setuptools/archive/refs/tags/v77.0.2.zip'
# tag not published to PyPI.
python
>>> import setuptools
>>> from distutils.command.sdist import sdist
>>> sdist.__mro__
(<class 'distutils.command.sdist.sdist'>, <class 'distutils.cmd.Command'>, <class 'object'>)
```

I don't know if there are the deeper unforeseen consequences of this change (other than problems similar to https://github.com/pypa/setuptools/issues/4902).

_Originally posted by @abravalheri in https://github.com/pypa/distutils/issues/345#issuecomment-2741407539_

abravalheri:
Fix #351
`Fix` pypa/setuptools#4902
`Fix` pypa/setuptools#4906
`Fix` pypa/setuptools#4908

Uses the approach described in https://github.com/pypa/distutils/pull/345#issuecomment-2741454564

### DIFF
--- a/distutils/command/__init__.py
+++ b/distutils/command/__init__.py
@@ -3,24 +3,27 @@
 Package containing implementation of all the standard Distutils
 commands."""
 
-from . import (
-    bdist,
-    bdist_dumb,
-    bdist_rpm,
-    build,
-    build_clib,
-    build_ext,
-    build_py,
-    build_scripts,
-    check,
-    clean,
-    install,
-    install_data,
-    install_headers,
-    install_lib,
-    install_scripts,
-    sdist,
-)
+from typing import TYPE_CHECKING
+
+if TYPE_CHECKING:
+    from . import (
+        bdist,
+        bdist_dumb,
+        bdist_rpm,
+        build,
+        build_clib,
+        build_ext,
+        build_py,
+        build_scripts,
+        check,
+        clean,
+        install,
+        install_data,
+        install_headers,
+        install_lib,
+        install_scripts,
+        sdist,
+    )
 
 __all__ = [
     'bdist',


abravalheri:
In https://github.com/pypa/distutils/pull/352/commits/81a8a61facffb9321ffa047da765bdf2805ff1a0 I am trying to add a quick regression test...

Not sure if that is that is the best approach (very possibly not).
So I will probably revert the commit once I see the CI working or not.

After it is merged we can add the test to the setuptools suite.

jaraco:
As discussed in #345, there's still some uncertainty about the effectiveness of this approach. Let's just do a simple revert and then explore more permanent solutions to the typing system separately.

PR #353, opened 2025-04-03
Always return an actual `bool` from ``has_*`` functions

Avasam:
Reduce changes from #343

Technically a user-facing change, but not sure how to present that in a changelog for setuptools, considering the changelog entry "Always return an actual `bool` from ``has_*`` functions" wouldn't apply to all setuptools' own methods.

### DIFF
--- a/distutils/dist.py
+++ b/distutils/dist.py
@@ -1024,25 +1024,25 @@ Common commands: (see '--help-commands' for more)
     # -- Distribution query methods ------------------------------------
 
     def has_pure_modules(self) -> bool:
-        return len(self.packages or self.py_modules or []) > 0
+        return bool(self.packages or self.py_modules)
 
     def has_ext_modules(self) -> bool:
-        return bool(self.ext_modules and len(self.ext_modules) > 0)
+        return bool(self.ext_modules)
 
     def has_c_libraries(self) -> bool:
-        return bool(self.libraries and len(self.libraries) > 0)
+        return bool(self.libraries)
 
     def has_modules(self) -> bool:
         return self.has_pure_modules() or self.has_ext_modules()
 
     def has_headers(self) -> bool:
-        return bool(self.headers and len(self.headers) > 0)
+        return bool(self.headers)
 
     def has_scripts(self) -> bool:
-        return bool(self.scripts and len(self.scripts) > 0)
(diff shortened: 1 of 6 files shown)


distutils/dist.py
```diff
@@ -1027,22 +1027,22 @@ def has_pure_modules(self) -> bool:
         return len(self.packages or self.py_modules or []) > 0
 
     def has_ext_modules(self) -> bool:
-        return self.ext_modules and len(self.ext_modules) > 0
+        return bool(self.ext_modules and len(self.ext_modules) > 0)
```
jaraco, line 1030: Agreed returning a proper bool makes sense; what do you think about these alternatives?

```suggestion
        return bool(self.ext_modules) and len(self.ext_modules) > 0
```


```suggestion
    @apply(bool)
    def has_ext_modules(self) -> bool
        return self.ext_modules and len(self.ext_modules) > 0
```

jaraco: And now that I look at these functions again, I realize they're way over-engineered. Collections already have boolean support.

```suggestion
        return bool(self.ext_modules)
```

Avasam: To answer your first comment:
1. I indiscriminately wrapped the entire expression. Which means I didn't have to call `bool` twice anywhere.
2. That `@apply` trick would work if the decorator is properly typed to forward its parameter return type. But you also have to change the declared return type of `` or type-checkers will correctly point out that you're returning a `list` instead of a `bool`.

jaraco:
After reflecting on this change for a bit, I realized I don't want to apply this technique generally.

One of the nice things about Python is the ability to use the boolean inference to make checks on objects that aren't themselves boolean. It's nice not to have to write `if bool(list_obj)` and just write `if list_obj`. I'd like to see the type checkers start supporting a SupportsBool protocol or similar... or allow any object to be used in place of a `bool`.

I'm guessing the typechecker authors have already considered this problem and realized that since Any python object can be treated as a boolean, there's little strictness in allowing a declared 'bool' to resolve from Any.

I do want to try avoid sprinkling `bool()` all over the place simply because the typecheckers are being nitpicky. The simple elegance of boolean inference would be lost, so I'll be resistant to changes like these in the future.

PR #356, opened 2025-04-18
Respect CXX when parsing linker parameters for UNIX c++ targets

### LINKED ISSUE #355: Linking cxx shared libraries may fail when CC != CXX
Noticed this while building greenlet, which uses c++.

Starting with https://github.com/pypa/distutils/commit/52cd70b8dd7f5c3405ab0d4a9719b4cb0493fe9e

The LDCXXSHARED variable gets read out of the sysconfig vars from the python build:

```
br-user@02db68569e9b:/tmp/tmp.SWYRHhln2t/buildroot$ /tmp/tmp.SWYRHhln2t/buildroot/output/host/bin/python3 -m sysconfig | grep LDC
	LDCXXSHARED = "/tmp/tmp.SWYRHhln2t/buildroot/output/host/bin/ccache /usr/bin/g++ -shared -L/tmp/tmp.SWYRHhln2t/buildroot/output/host/lib -Wl,-rpath,/tmp/tmp.SWYRHhln2t/buildroot/output/host/lib -Wl,--enable-new-dtags"

```

This feeds the value of `linker_so_cxx`:

```python
        compiler.set_executables(
            preprocessor=cpp,
            compiler=cc_cmd,
            compiler_so=cc_cmd + ' ' + ccshared,
            compiler_cxx=cxx_cmd,
            compiler_so_cxx=cxx_cmd + ' ' + ccshared,
            linker_so=ldshared,
            linker_so_cxx=ldcxxshared,
            linker_exe=cc,
            linker_exe_cxx=cxx,
            archiver=archiver,
        )
```

When linking is performed:
https://github.com/pypa/distutils/blob/24bd3179ba784afc00aae52101df064b826be0c0/distutils/compilers/C/unix.py#L289-L295

```
linker=['/tmp/tmp.SWYRHhln2t/buildroot/output/host/bin/ccache', '/usr/bin/g++', '-shared', '-L/tmp/tmp.SWYRHhln2t/buildroot/output/host/lib', '-Wl,-rpath,/tmp/tmp.SWYRHhln2t/buildroot/output/host/lib', '-Wl,--enable-new-dtags', '-L/tmp/tmp.SWYRHhln2t/buildroot/output/host/lib', '-Wl,-rpath,/tmp/tmp.SWYRHhln2t/buildroot/output/host/lib', '-O2', '-I/tmp/tmp.SWYRHhln2t/buildroot/output/host/include', '-I/tmp/tmp.SWYRHhln2t/buildroot/output/host/include']

```

However, this ultimately fails:

```
/tmp/tmp.SWYRHhln2t/buildroot/output/host/bin/ccache /usr/bin/g++ -O2 -I/tmp/tmp.SWYRHhln2t/buildroot/output/host/include -I/tmp/tmp.SWYRHhln2t/buildroot/output/host/include /usr/bin/g++ -shared -L/tmp/tmp.SWYRHhln2t/buildroot/output/host/lib -Wl,-rpath,/tmp/tmp.SWYRHhln2t/buildroot/output/host/lib -Wl,--enable-new-dtags -L/tmp/tmp.SWYRHhln2t/buildroot/output/host/lib -Wl,-rpath,/tmp/tmp.SWYRHhln2t/buildroot/output/host/lib -O2 -I/tmp/tmp.SWYRHhln2t/buildroot/output/host/include -I/tmp/tmp.SWYRHhln2t/buildroot/output/host/include -g build/temp.linux-x86_64-cpython-313/src/greenlet/greenlet.o -L/tmp/tmp.SWYRHhln2t/buildroot/output/host/lib -o build/lib.linux-x86_64-cpython-313/greenlet/_greenlet.cpython-313-x86_64-linux-gnu.so
/usr/bin/ld: cannot use executable file '/usr/bin/g++' as input to a link
collect2: error: ld returned 1 exit status
error: command '/tmp/tmp.SWYRHhln2t/buildroot/output/host/bin/ccache' failed with exit code 1

```

```
linker=['/tmp/tmp.SWYRHhln2t/buildroot/output/host/bin/ccache', '/usr/bin/g++', '-O2', '-I/tmp/tmp.SWYRHhln2t/buildroot/output/host/include', '-I/tmp/tmp.SWYRHhln2t/buildroot/output/host/include', '/usr/bin/g++', '-shared', '-L/tmp/tmp.SWYRHhln2t/buildroot/output/host/lib', '-Wl,-rpath,/tmp/tmp.SWYRHhln2t/buildroot/output/host/lib', '-Wl,--enable-new-dtags', '-L/tmp/tmp.SWYRHhln2t/buildroot/output/host/lib', '-Wl,-rpath,/tmp/tmp.SWYRHhln2t/buildroot/output/host/lib', '-O2', '-I/tmp/tmp.SWYRHhln2t/buildroot/output/host/include', '-I/tmp/tmp.SWYRHhln2t/buildroot/output/host/include']
ld_args=['-g', 'build/temp.linux-x86_64-cpython-313/src/greenlet/greenlet.o', '-L/tmp/tmp.SWYRHhln2t/buildroot/output/host/lib', '-o', 'build/lib.linux-x86_64-cpython-313/greenlet/_greenlet.cpython-313-x86_64-linux-gnu.so']

```

The linking command has a spurious `/usr/bin/g++` in the invocation, causing problems.

This looks to be because `LDCXXSHARED` is always generated using `CXX` in the CPython build:
https://github.com/python/cpython/blob/e42bda9441119c7952ada23e88e6f4ab79df86c2/configure.ac#L3469

However, when parsing this out in `unix.py`:
https://github.com/pypa/distutils/blob/24bd3179ba784afc00aae52101df064b826be0c0/distutils/compilers/C/unix.py#L301-L304

`self.linker_exe` is used instead of `self.linker_exe_cxx`. `self.linker_exe` is driven by `CC` where as `self.linker_exe_cxx` is driven by `CXX`.

```
br-user@02db68569e9b:/tmp/tmp.SWYRHhln2t/buildroot$ /tmp/tmp.SWYRHhln2t/buildroot/output/host/bin/python3 -m sysconfig | egrep "(CXX|CC) =" 
	CC = "/tmp/tmp.SWYRHhln2t/buildroot/output/host/bin/ccache /usr/bin/gcc"
	CXX = "/tmp/tmp.SWYRHhln2t/buildroot/output/host/bin/ccache /usr/bin/g++"
	LINKCC = "/tmp/tmp.SWYRHhln2t/buildroot/output/host/bin/ccache /usr/bin/gcc"
```

Since the LDCXXSHARED value is prefixed with CXX and not CC, only the first item from the list is dropped via `_linker_params`, which is the path to ccache, and the spurious `/usr/bin/g++` ref is left around.

So, I think maybe the `linker_exe_cxx` attribute should be used instead so the proper prefix is parsed out of the flag:

```python
                if target_lang == "c++" and self.compiler_cxx:
                    env, linker_ne = _split_env(linker)
                    aix, linker_na = _split_aix(linker_ne)
                    _, compiler_cxx_ne = _split_env(self.compiler_cxx)
                    _, linker_exe_ne = _split_env(self.linker_exe_cxx)
```

There may also be issues with c++ executables:
https://github.com/pypa/distutils/blob/24bd3179ba784afc00aae52101df064b826be0c0/distutils/compilers/C/unix.py#L290-L291

There is no branch to use `self.linker_exe_cxx` when `target_lang == "c++"`

vfazio:
Previously, when parsing linker parameters for C++ targets, the CC
variable was used to determine what the "prefix" of the command was in
order to determine what the linker arguments were.

If the value of LDCXXSHARED did not match CC, the first argument would
be dropped as it was assumed to be the linker command.

However, if the command was a wrapper, such as ccache, it could lead to
compile problems as the generated command would be incorrect.

In the following scenario:
  LDCXXSHARED="ccache g++ -shared -Wl,--enable-new-dtags"
  CC="ccache gcc"
  CXX="ccache g++"

The command would be incorrectly parsed to:
  "ccache g++ g++ -shared -Wl,--enable-new-dtags"

Now, the CXX value is used to improve the chances of parsing the linker
arguments correctly to generate:
  "ccache g++ -shared -Wl,--enable-new-dtags"

LDCXXSHARED and CXX still need to be in sync either in the environment
or within the sysconfig variables in the CPython build for parsing to
work correctly.

The CXX value is now also respected when linking executable binaries.

closes #355

### DIFF
--- a/distutils/compilers/C/tests/test_unix.py
+++ b/distutils/compilers/C/tests/test_unix.py
@@ -260,7 +260,7 @@ class TestUnixCCompiler(support.TempdirManager):
         def gcvs(*args, _orig=sysconfig.get_config_vars):
             if args:
                 return list(map(sysconfig.get_config_var, args))
-            return _orig()
+            return _orig()  # pragma: no cover
 
         sysconfig.get_config_var = gcv
         sysconfig.get_config_vars = gcvs
@@ -270,6 +270,10 @@ class TestUnixCCompiler(support.TempdirManager):
             mock.patch.object(self.cc, 'mkpath', return_value=None),
             EnvironmentVarGuard() as env,
         ):
+            # override environment overrides in case they're specified by CI
+            del env['CXX']
+            del env['LDCXXSHARED']
+
             sysconfig.customize_compiler(self.cc)
             assert self.cc.linker_so_cxx[0:2] == ['ccache', 'g++-4.2']
             assert self.cc.linker_exe_cxx[0:2] == ['ccache', 'g++-4.2']
@@ -280,7 +284,7 @@ class TestUnixCCompiler(support.TempdirManager):
 
             self.cc.link_executable([], 'a.out', target_lang='c++')
             call_args = mock_spawn.call_args[0][0]
-            expected = ['ccache', 'g++-4.2', '-o', 'a.out']
+            expected = ['ccache', 'g++-4.2', '-o', self.cc.executable_filename('a.out')]
             assert call_args[:4] == expected
 
             env['LDCXXSHARED'] = 'wrapper g++-4.2 -bundle -undefined dynamic_lookup'
@@ -295,7 +299,12 @@ class TestUnixCCompiler(support.TempdirManager):
 
             self.cc.link_executable([], 'a.out', target_lang='c++')
             call_args = mock_spawn.call_args[0][0]
-            expected = ['wrapper', 'g++-4.2', '-o', 'a.out']
+            expected = [
+                'wrapper',
+                'g++-4.2',
+                '-o',
+                self.cc.executable_filename('a.out'),
+            ]
             assert call_args[:4] == expected
 
     @pytest.mark.skipif('platform.system == "Windows"')
--- a/distutils/compilers/C/tests/test_unix.py
+++ b/distutils/compilers/C/tests/test_unix.py
@@ -245,6 +245,60 @@ class TestUnixCCompiler(support.TempdirManager):
         assert self.cc.linker_so[0] == 'my_cc'
 
     @pytest.mark.skipif('platform.system == "Windows"')
+    def test_cxx_commands_used_are_correct(self):
+        def gcv(v):
+            if v == 'LDSHARED':
+                return 'ccache gcc-4.2 -bundle -undefined dynamic_lookup'
+            elif v == 'LDCXXSHARED':
+                return 'ccache g++-4.2 -bundle -undefined dynamic_lookup'
+            elif v == 'CXX':
+                return 'ccache g++-4.2'
+            elif v == 'CC':
+                return 'ccache gcc-4.2'
+            return ''
+
+        def gcvs(*args, _orig=sysconfig.get_config_vars):
+            if args:
+                return list(map(sysconfig.get_config_var, args))
+            return _orig()
+
+        sysconfig.get_config_var = gcv
+        sysconfig.get_config_vars = gcvs
+        with (
+            mock.patch.object(self.cc, 'spawn', return_value=None) as mock_spawn,
+            mock.patch.object(self.cc, '_need_link', return_value=True),
+            mock.patch.object(self.cc, 'mkpath', return_value=None),
+            EnvironmentVarGuard() as env,
+        ):
+            sysconfig.customize_compiler(self.cc)
+            assert self.cc.linker_so_cxx[0:2] == ['ccache', 'g++-4.2']
+            assert self.cc.linker_exe_cxx[0:2] == ['ccache', 'g++-4.2']
+            self.cc.link(None, [], 'a.out', target_lang='c++')
+            call_args = mock_spawn.call_args[0][0]
+            expected = ['ccache', 'g++-4.2', '-bundle', '-undefined', 'dynamic_lookup']
+            assert call_args[:5] == expected
+
+            self.cc.link_executable([], 'a.out', target_lang='c++')
+            call_args = mock_spawn.call_args[0][0]
+            expected = ['ccache', 'g++-4.2', '-o', 'a.out']
+            assert call_args[:4] == expected
+
+            env['LDCXXSHARED'] = 'wrapper g++-4.2 -bundle -undefined dynamic_lookup'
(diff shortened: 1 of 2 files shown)


jaraco:
This is excellent. It makes the code even more symmetric and captures the failure clearly. The thorough analysis in the reported issue gives me high confidence this is the right fix. Thank you for everything.

vfazio:
> This is excellent. It makes the code even more symmetric and captures the failure clearly. The thorough analysis in the reported issue gives me high confidence this is the right fix. Thank you for everything.

derp, except it looks like its breaking mingwin builds.. i'll probably need help debugging that

jaraco:
> derp, except it looks like its breaking mingwin builds.. i'll probably need help debugging that

No problem. Please take a look and if the fix isn't obvious, feel free to tag in `@lazka` with relevant questions.

vfazio:
Ok, may be ready for another run. I don't have a mingwin box to putz with but my guess is it's the CXX environment variable that is being injected via the CI pipeline. I guess we'll find out what else I missed after this

vfazio:
@jaraco do you mind kicking off another pipeline at your convenience? Hopefully, if everything succeeds, I'd like to backport this to fix some build issues we're seeing in Buildroot until it gets pulled into a new version of setuptools.

Thanks!

jaraco:
> I'd like to backport this to fix some build issues we're seeing in Buildroot until it gets pulled into a new version of setuptools.

I should be able to get it rolled out in Setuptools in short order, if that helps.

vfazio:
> > I'd like to backport this to fix some build issues we're seeing in Buildroot until it gets pulled into a new version of setuptools.
> 
> I should be able to get it rolled out in Setuptools in short order, if that helps.

That would be really helpful! Thanks!

I'm always a bit nervous when touching important packages due to the chance to ripple and cripple, so hopefully this changeset is concise, reasonable, and doesn't end up causing headaches for users.

jaraco:
> hopefully this changeset is concise, reasonable, and doesn't end up causing headaches for users.

Exactly. And most importantly, it adds a new test case that will prevent from regression, so we're continuing to iterate toward perfection. Thanks for the contrib!

jaraco:
Setuptools fix pending in https://github.com/pypa/setuptools/pull/4959.

PR #362, opened 2025-05-04
mypy: Remove all-fixed `func-returns-value` and `has-type` from exclusions

Avasam:
~~This would "fix itself" with https://github.com/pypa/distutils/pull/335~~

### DIFF
--- a/distutils/dir_util.py
+++ b/distutils/dir_util.py
@@ -57,10 +57,11 @@ def mkpath(name: pathlib.Path, mode=0o777, verbose=True, dry_run=False) -> None:
     if verbose and not name.is_dir():
         log.info("creating %s", name)
 
-    try:
-        dry_run or name.mkdir(mode=mode, parents=True, exist_ok=True)
-    except OSError as exc:
-        raise DistutilsFileError(f"could not create '{name}': {exc.args[-1]}")
+    if not dry_run:
+        try:
+            name.mkdir(mode=mode, parents=True, exist_ok=True)
+        except OSError as exc:
+            raise DistutilsFileError(f"could not create '{name}': {exc.args[-1]}")
 
 
 @mkpath.register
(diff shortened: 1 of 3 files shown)


distutils/dir_util.py
```diff
@@ -57,10 +57,11 @@ def mkpath(name: pathlib.Path, mode=0o777, verbose=True, dry_run=False) -> None:
...
 
-    try:
-        dry_run or name.mkdir(mode=mode, parents=True, exist_ok=True)
-    except OSError as exc:
-        raise DistutilsFileError(f"could not create '{name}': {exc.args[-1]}")
+    if not dry_run:
+        try:
+            name.mkdir(mode=mode, parents=True, exist_ok=True)
+        except OSError as exc:
+            raise DistutilsFileError(f"could not create '{name}': {exc.args[-1]}")
```
Avasam: https://github.com/pypa/distutils/pull/343/files#r2440539508
> This one could be ignored on the line itself without being unsafe.
> 
> It's an unused expression, both mypy and pyright warn here because of the lack of assignement, it "looks like" a possible mistake because the result of the expression is not used for anything. But it's just a conditional execution pattern.
> 
> It's definitely more of a lint than a type error.
> 
> * [microsoft/pylance-release@`main`/docs/diagnostics/reportUnusedExpression.md](https://github.com/microsoft/pylance-release/blob/main/docs/diagnostics/reportUnusedExpression.md)
> * [microsoft/pyright@`main`/docs/configuration.md#user-content-reportunusedexpression](https://github.com/microsoft/pyright/blob/main/docs/configuration.md#user-content-reportunusedexpression)
> * [mypy.readthedocs.io/en/latest/error_code_list.html#code-func-returns-value](https://mypy.readthedocs.io/en/latest/error_code_list.html#code-func-returns-value)
> 
> FWIW, #334 / #335 would remove the condition anyway.

PR #363, opened 2025-05-04
mypy: Fix all index

Avasam:


distutils/util.py
```diff
@@ -151,7 +151,9 @@ def change_root(
         if not os.path.isabs(pathname):
             return os.path.join(new_root, pathname)
         else:
-            return os.path.join(new_root, pathname[1:])
+            # type-ignore: This makes absolutes os.Pathlike unsupported in this branch.
+            # Either this or we don't support bytes-based paths, or we complexify this branch.
+            return os.path.join(new_root, pathname[1:])  # type: ignore[index]
```
Avasam, line 156: There's a decision to be taken here.
1. Don't support `PathLike` in this branch (current behaviour)
2. Don't support `bytes | PathLike[bytes]` in this branch
```suggestion
            # type-ignore: This makes absolutes os.Pathlike unsupported in this branch.
            # Either this or we don't support bytes-based paths, or we complexify this branch.
            return os.path.join(new_root, str(pathname)[1:])
```
3. Support both, requires extra checking.

jaraco: I like option 2. IMHO, `bytes` should be avoided for pathnames. What's the likelihood that bytes are used in the wild?

Avasam: > What's the likelihood that bytes are used in the wild?

I'd say very low, I also agree with preferring support for PathLike without bloating the complexity.

Avasam: Done. It's almost an implementation detail, is it worth mentioning that `change_root` should now support absolute `PathLike` (most likely a `path.Path`) on posix?  (it already did for relative paths and nt)

distutils/sysconfig.py
```diff
@@ -313,16 +313,22 @@ def customize_compiler(compiler: CCompiler) -> None:
...
-            'LDSHARED',
-            'LDCXXSHARED',
-            'SHLIB_SUFFIX',
-            'AR',
-            'ARFLAGS',
+        ) = cast(
+            # These being anything else than str would result in errors later on, but see:
+            # - https://github.com/pypa/distutils/pull/366#discussion_r2440251813
+            # - https://github.com/pypa/distutils/pull/343#discussion_r2440210538
+            "tuple[str, ...]",
```
jaraco: Let's implement this cast as a function so it can document its purpose and act more concisely here. E.g. create `_get_config_vars_safe` that returns the `tuple[str ,...]`.

I wonder if we shouldn't just adapt `get_config_vars` to enforce that return type.

jaraco: (exploring with Claude...)

Avasam: > I wonder if we shouldn't just adapt `get_config_vars` to enforce that return type.

I need to check again why, but I vaguely remember there being a current edge case where one value could be an `int`

Edit: Yeah the comment you added mentions the possibility of `int` type.
I think it's reasonable to have extracted the "we know what we're doing here, only using the `str` possible types, `int` would fail anyway"

PR #366, opened 2025-05-04
mypy: Fix all union-attr

Avasam:


distutils/command/build_py.py
```diff
@@ -360,7 +360,7 @@ def build_modules(self) -> None:
             self.build_module(module, module_file, package)
 
     def build_packages(self) -> None:
-        for package in self.packages:
+        for package in self.packages or ():
```
Avasam, line 363: This is a change in behaviour as it would graciously do nothing instead of raising if `self.packages` is still `None`.

distutils/util.py
```diff
@@ -232,14 +232,9 @@ def grok_environment_error(exc: object, prefix: str = "error: ") -> str:
...
-
-
-def _init_regex():
-    global _wordchars_re, _squote_re, _dquote_re
-    _wordchars_re = re.compile(rf'[^\\\'\"{string.whitespace} ]*')
-    _squote_re = re.compile(r"'(?:[^'\\]|\\.)*'")
-    _dquote_re = re.compile(r'"(?:[^"\\]|\\.)*"')
+_wordchars_re = re.compile(rf'[^\\\'\"{string.whitespace} ]*')
+_squote_re = re.compile(r"'(?:[^'\\]|\\.)*'")
+_dquote_re = re.compile(r'"(?:[^"\\]|\\.)*"')
```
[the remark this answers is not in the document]
jaraco: Agreed. Looks like premature optimization. Even the `re.compile` calls are probably premature. Probably could have just used simple strings and compiled on demand and relied on caching in the re module to avoid duplicate work, but this approach is clean and simple enough.

distutils/command/build_py.py
jaraco: Presumably `self.packages` is expected not to be None at this point. I'd rather see it enforced that it's not None by the time it reaches this point. That's probably more work than it's worth, so for a local fix, let's just do an explicit check (`if self.packages is None` raise a clear error).

distutils/sysconfig.py
```diff
@@ -324,6 +324,16 @@ def customize_compiler(compiler: CCompiler) -> None:
...
+        assert isinstance(cc, str)
+        assert isinstance(cxx, str)
+        assert isinstance(cflags, str)
+        assert isinstance(ccshared, str)
+        assert isinstance(ldshared, str)
+        assert isinstance(ldcxxshared, str)
+        assert isinstance(shlib_suffix, str)
+        assert isinstance(ar_flags, str)
+        ar = os.environ.get('AR', ar)
+        assert isinstance(ar, str)
```
jaraco: I mentioned in another PR - I'm not sure these changes are stable. I'm worried there are edge cases that would begin to fail with this change. Let's be confident that this change doesn't break anything (and consider using casts to retain the current behavior if appropriate).

jaraco (approved):

distutils/command/build_py.py
Avasam: Feel free to update the new message to your likings !

distutils/sysconfig.py
Avasam: Resolved !

distutils/cmd.py
```diff
@@ -314,7 +314,8 @@ def get_finalized_command(self, command: str, create: bool = True) -> Command:
         'command', call its 'ensure_finalized()' method, and return the
         finalized command object.
         """
-        cmd_obj = self.distribution.get_command_obj(command, create)
+        # TODO: Raise a more descriptive error when create=False or cmd_obj is None ?
+        cmd_obj = cast(Command, self.distribution.get_command_obj(command, True))
```
Avasam: Probably better handled by https://github.com/pypa/distutils/pull/377

PR #368, opened 2025-05-04
mypy: fix all assignment

Avasam:


distutils/fancy_getopt.py
```diff
@@ -260,21 +261,21 @@ def getopt(self, args: Sequence[str] | None = None, object=None):  # noqa: C901
...
             # The only repeating option at the moment is 'verbose'.
             # It has a negative option -q quiet, which should set verbose = False.
-            if val and self.repeat.get(attr) is not None:
-                val = getattr(object, attr, 0) + 1
-            setattr(object, attr, val)
-            self.option_order.append((opt, val))
+            if value and self.repeat.get(attr) is not None:
+                value = getattr(object, attr, 0) + 1
+            setattr(object, attr, value)
+            self.option_order.append((opt, value))
```
Avasam: See discussion in https://github.com/pypa/distutils/pull/343/files#r2440199822

distutils/sysconfig.py
```diff
@@ -324,6 +324,16 @@ def customize_compiler(compiler: CCompiler) -> None:
...
+        assert isinstance(cc, str)
+        assert isinstance(cxx, str)
+        assert isinstance(cflags, str)
+        assert isinstance(ccshared, str)
+        assert isinstance(ldshared, str)
+        assert isinstance(ldcxxshared, str)
+        assert isinstance(shlib_suffix, str)
+        assert isinstance(ar_flags, str)
+        ar = os.environ.get('AR', ar)
+        assert isinstance(ar, str)
```
Avasam: https://github.com/pypa/distutils/pull/366#discussion_r2440251813 
> I'm not sure these changes are stable. I'm worried there are edge cases that would begin to fail with this change. Let's be confident that this change doesn't break anything (and consider using casts to retain the current behavior if appropriate).

https://github.com/pypa/distutils/pull/343#discussion_r2440210538
> I'm not sure these assertions will always pass. There have been some bugs around these being undefined in unusual environments. Maybe it makes sense to catch these conditions early. Do we have good reason to believe that non-string values returned here would never be viable (would be breaking anyway)?

distutils/command/build_ext.py
```diff
@@ -433,7 +433,7 @@ def check_extensions_list(self, extensions) -> None:  # noqa: C901
                     setattr(ext, key, val)
 
             # Medium-easy stuff: same syntax/semantics, different names.
-            ext.runtime_library_dirs = build_info.get('rpath')
+            ext.runtime_library_dirs = build_info.get('rpath') or []
```
Avasam, line 444: Runtime change: avoid assigning `None` to `runtime_library_dirs` (I didn't fully investigate whether that *should* even be possible at runtime or just an artefact of code that is too dynamic to be type-safe)

distutils/command/build_ext.py
```diff
@@ -110,28 +110,28 @@ class build_ext(Command):
...
-        self.library_dirs = None
-        self.rpath = None
+        self.libraries: list[str] = None  # type: ignore[assignment] # Should always be set in finalize_options
+        self.library_dirs: list[str] = None  # type: ignore[assignment] # Should always be set in finalize_options
+        self.rpath: list[str] = None  # type: ignore[assignment] # Should always be set in finalize_options
         self.link_objects = None
         self.debug = None
-        self.force: bool = None  # Should always be set in finalize_options
-        self.compiler = None
+        self.force: bool = None  # type: ignore[assignment] # Should always be set in finalize_options
```
Avasam: I'm not seeing where `build_ext.force` is being coerced to `bool`

distutils/command/build_ext.py
```diff
@@ -110,28 +110,28 @@ class build_ext(Command):
...
+        self.compiler: CCompiler | None = None
```
Avasam: Tests failing because there's still 3 spots where code assumes `build_ext.compiler` can't be `None`. Makes sense to me, but I don't immediately see where we ensure it can't be `None` leaving `finalize_options`.

distutils/tests/unix_compat.py
```diff
@@ -1,12 +1,18 @@
...
+import pytest
 
+grp: ModuleType | None = None
+pwd: ModuleType | None = None
 try:
     import grp
     import pwd
 except ImportError:
-    grp = pwd = None
+    pass
```
jaraco: It's more than a little annoying that adding type hints causes the imports to move from the import section of the module to the post-imports section. I wonder what happens if we just declare the type hints, but leave the implementation as it was.

jaraco: I replaced it with:

```diff
diff --git a/distutils/tests/unix_compat.py b/distutils/tests/unix_compat.py
index af020ee64..625a9a4b7 100644
--- a/distutils/tests/unix_compat.py
+++ b/distutils/tests/unix_compat.py
@@ -5,13 +5,13 @@ from types import ModuleType
 
 import pytest
 
-grp: ModuleType | None = None
-pwd: ModuleType | None = None
+grp: ModuleType | None
+pwd: ModuleType | None
 try:
     import grp
     import pwd
 except ImportError:
-    pass
+    grp = pwd = None
 
 
 UNIX_ID_SUPPORT = grp and pwd
```

But ruff still thinks the `import pytest` belongs above, so it's the presence of the type annotations that causes the switch out of the import section.

jaraco: Let's use `contextlib.suppress(ImportError)` instead of `except/pass`.

distutils/archive_util.py
```diff
@@ -6,29 +6,37 @@
...
 from .errors import DistutilsExecError
 from .spawn import spawn
 
+zipfile: ModuleType | None = None
 try:
-    from pwd import getpwnam
+    import zipfile
 except ImportError:
-    getpwnam = None
+    pass
```
jaraco: Use contextlib.suppress

Avasam: For the record, using `contextlib.suppress` only saves a single line and is much slower. So it would be a purely cosmetic change.

distutils/archive_util.py
```diff
@@ -6,29 +6,37 @@
...
+else:
+    try:
+        from pwd import getpwnam
+    except ImportError:
+        getpwnam = None
+
+    try:
+        from grp import getgrnam
+    except ImportError:
+        getgrnam = None
```
jaraco: This pattern looks very different from the other conditional imports. Why not use the same pattern as zipfile above and the tests.unix_compat module, and then use `getgrnam = getattr(grp, 'getgrnam')`? Then leave a comment to the effect of "type checkers make conditional attribute loading complicated; ref python/mypy#1393".

distutils/fancy_getopt.py
jaraco: Let's implement the alternative suggested there (cast).

distutils/command/build_ext.py
jaraco: Sounds good. We can keep this; I prefer using degenerate forms to nullable types.

distutils/fancy_getopt.py
Avasam: I need to try this out again with mypy's newer re-assignement rules (may require config changes)

distutils/command/build_ext.py
jaraco: This value can also be a `str` when it's specified from the command line [via set_undefined_options](https://github.com/pypa/distutils/blob/5d4dcbd8d9824a80c04da902fd601a9a8abd1c01/distutils/cmd.py#L286). And it's only [converted from a str to a CCompiler in run](https://github.com/pypa/distutils/blob/5d4dcbd8d9824a80c04da902fd601a9a8abd1c01/distutils/command/build_ext.py#L326-L331).

We just need to figure out how to tell the type system that this attribute goes through three different types in three different phases, or to just ignore this attribute.

Long term, we probably want to avoid mutating the meaning of `self.compiler`, possibly by passing the compiler instead of storing it as an instance attribute, but for now, let's focus on getting the annotations to work with the current implementation.

jaraco:
I'm going to see what Claude can do with the suggestions. Standby...

distutils/archive_util.py
Avasam: > Why not use the same pattern as zipfile above and the tests.unix_compat module, and then use getgrnam = getattr(grp, 'getgrnam')?

Just because I'm always attempting to reduce runtime changes, and letting you decide on concessions :)
Changing to `import grp` should work like the other modules.

---

Otherwise,

The clean type-checked look, would've looked like this:

```py
if sys.platform == "win32":
    getpwnam = None
    getgrnam = None
else:
    from grp import getgrnam
    from pwd import getpwnam
```

But that technically changes the actual logic (from "missing import" to "OS check assuming those imports will be missing").

You also can't pre-declare the type because the return types comes from the modules you can't import on Windows in the first place. So if you wanna do that, you HAVE to use a TYPE_CHECKING block:

```py
# Impossible without special TYPE_CHECKING block because struct_passwd also comes from pwd
getpwnam: Callable[[str], struct_passwd]
with contextlib.suppress(ImportError):
    from pwd import getpwnam
```

distutils/command/build_ext.py
Avasam: > Long term, we probably want to avoid mutating the meaning of `self.compiler`, possibly by passing the compiler instead of storing it as an instance attribute, but for now, let's focus on getting the annotations to work with the current implementation.

Definitly, in pywin32 I tried having this attribute be typed and ended up more annoyed than anything that it could be a `str` in 1 place, then a class everywhere else.
It was also a source of broken MinGW compilation for pywin32 for many years because the wrong comparison was done at the wrong time. Making it always false.

As annoying as it can be, I think type correctness will be more important here, as it'll make the footgun obvious.

distutils/archive_util.py
jaraco: I'd prefer if it wasn't branchy at all; if we could simply: `maybe_import zipfile` or `zipfile = maybe_import('zipfile')`, encapsulating this very common pattern. The former would require a syntax change. The latter could be done in a third-party library. Even better would be for the stdlib to provide these modules unconditional on platform (having a degenerate or compatibility version on platforms where the functionality is unavailable). For now, let's use the most concise form available and we can optimize for performance later if needed.

distutils/command/build_ext.py
```diff
@@ -565,7 +569,10 @@ def build_extension(self, ext) -> None:
         for undef in ext.undef_macros:
             macros.append((undef,))
 
-        objects = self.compiler.compile(
+        # By the time extensions are built, run() has replaced the compiler
+        # name with a CCompiler instance.
+        compiler = cast(CCompiler, self.compiler)
```
Avasam: Imo casting here feels wrong. type-checkers are absolutely correct that this is not type-safe, and not because of a limitation or a weird 2-step initialization (since this is coerced in `run` rather than `finalize_options`), or due to a manual check they don't understand.

I fixed this on my end by simply explicitly rethrowing with a more descriptive error message. I merged the conflict, but you can take whichever version you prefer.

distutils/archive_util.py
Avasam: These are 1 time import anyway. Hence I didn't really push back. Just had to *at least* mention it ^^

distutils/command/build_ext.py
jaraco: I'll keep your approach, but use a more concise `assert isinstance`.

Avasam:
@jaraco I believe all my questions and concerns have been resolved. I'm personally happy with the current state of this PR.

PR #370, opened 2025-05-06
msvc compiler: Re-sort sources to compile Message Compiler files (`.mc`) first

### LINKED ISSUE #4986: [FR] When sorting sources, move Message Compiler (`.mc`) files first
### What's the problem this feature will solve?

This is re-opening https://github.com/python/cpython/issues/86175 for setuptools.

If I'm not mistaken, https://github.com/python/cpython/pull/12341 is responsible for sorting sources in distutils to enable reproducible builds.

But since then, pywin32 has had to manually re-sort the sources by doing this:

```py
from distutils import ccompiler
from distutils._msvccompiler import MSVCCompiler

def my_new_compiler(**kw):
    if "compiler" in kw and kw["compiler"] in (None, "msvc"):
        return my_compiler()
    return orig_new_compiler(**kw)

# No way to cleanly wedge our compiler sub-class in.
orig_new_compiler = ccompiler.new_compiler
ccompiler.new_compiler = my_new_compiler

# This import has to be delayed to AFTER the compiler hack
from setuptools.command.build_ext import build_ext  # noqa: E402

class my_compiler(MSVCCompiler):
    # Work around bpo-36302/bpo-42009 - it sorts sources but this breaks
    # support for building .mc files etc :(
    def compile(self, sources, **kwargs) -> list[str]:
        # re-sort the list of source files but ensure all .mc files come first.
        def key_reverse_mc(a):
            b, e = os.path.splitext(a)
            e = "" if e == ".mc" else e
            return (e, b)

        sources = sorted(sources, key=key_reverse_mc)
        return MSVCCompiler.compile(self, sources, **kwargs)

    # [...]
```

If pywin32 didn't have to do this "compiler patching", I'm pretty sure I could remove this hack and 2 deprecated distutils imports

### Describe the solution you'd like

I'd like setuptools (distutils) to sort sources in such a way that Message Compiler (`.mc`) files come first.

### Alternative Solutions

(see the current workaround posted above)

### Additional context

- https://github.com/mhammond/pywin32/blob/main/setup.py
- https://github.com/pypa/setuptools/issues/4968

### Code of Conduct

- [x] I agree to follow the PSF Code of Conduct

Avasam:
Ensures that Message Compiler files come first.

Closes https://github.com/pypa/setuptools/issues/4986

Fixes pywin32's needs for hacking a custom MSVC compiler class.

### DIFF
--- a/distutils/compilers/C/msvc.py
+++ b/distutils/compilers/C/msvc.py
@@ -18,7 +18,8 @@ import os
 import subprocess
 import unittest.mock as mock
 import warnings
-from collections.abc import Iterable
+from collections.abc import Iterable, Sequence
+from pathlib import Path
 
 with contextlib.suppress(ImportError):
     import winreg
@@ -371,7 +372,7 @@ class Compiler(base.Compiler):
 
     def compile(  # noqa: C901
         self,
-        sources,
+        sources: Sequence[str | os.PathLike[str]],
         output_dir=None,
         macros=None,
         include_dirs=None,
@@ -379,9 +380,16 @@ class Compiler(base.Compiler):
         extra_preargs=None,
         extra_postargs=None,
         depends=None,
-    ):
+    ) -> list[str]:
         if not self.initialized:
             self.initialize()
+
+        # Move .mc files to the start of the list, otherwise keep the same order
+        # See pypa/setuptools#4986
+        sources = sorted(
+            sources, key=lambda source: Path(source).suffix not in self._mc_extensions
+        )
+
         compile_info = self._setup_compile(
             output_dir, macros, include_dirs, sources, depends, extra_postargs
         )
--- a/distutils/compilers/C/msvc.py
+++ b/distutils/compilers/C/msvc.py
@@ -17,7 +17,7 @@ import contextlib
 import os
 import subprocess
 import tempfile
-from collections.abc import Iterable, Iterator
+from collections.abc import Iterable, Iterator, Sequence
 from pathlib import Path
 from typing import ClassVar
 
@@ -431,7 +431,7 @@ class Compiler(base.Compiler):
 
     def compile(  # noqa: C901
         self,
-        sources,
+        sources: Sequence[str | os.PathLike[str]],
         output_dir=None,
         macros=None,
         include_dirs=None,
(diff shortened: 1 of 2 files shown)


distutils/compilers/C/msvc.py
```diff
@@ -371,17 +371,24 @@ def out_extensions(self) -> dict[str, str]:
...
         depends=None,
-    ):
+    ) -> list[str]:
         if not self.initialized:
             self.initialize()
+
+        # Re-sort the list of sources so that .mc files come first
+        # See pypa/setuptools#4986
+        sources = sorted(
+            sources, key=lambda source: (not str(source).endswith(".mc"), source)
```
Avasam, line 389: A tiny bit more complex, but I could look for the extensions in `_mc_extensions` instead of "hardocding" `.mc`.

I could also extract this to a function to better comment what they is doing:
The "key" is a tuple to do grouped sorting. The first element is `0` for mc files, which is sorted before `1` for everything else.
The second element is the original path to keep the same sorting as in `build_ext`.

Avasam:
I guess unfortunately coverage is gonna have to go down since this is only tested on Windows.

distutils/compilers/C/msvc.py
```diff
@@ -371,17 +371,22 @@ def out_extensions(self) -> dict[str, str]:
...
         extra_postargs=None,
         depends=None,
-    ):
+    ) -> list[str]:
         if not self.initialized:
             self.initialize()
+
+        # Move .mc files to the start of the list, otherwise keep the same order
+        # See pypa/setuptools#4986
+        sources = sorted(sources, key=lambda source: not str(source).endswith(".mc"))
```
Avasam, line 388: I could also look for the extensions in `_mc_extensions` instead of "hardocding" `.mc`.

Avasam:
Copying @raedrizqie 's comment here as well as my answer:

https://github.com/pypa/setuptools/issues/4986#issuecomment-3022026680
> Thanks @Avasam for this PR. Could you please made the sorting in `build_ext.build_extension`, so that MinGW could also benefit from it?

https://github.com/pypa/setuptools/issues/4986#issuecomment-3024626616
> I tried to only apply the change to compilers that would be relevant to building MC files. I'm not too familiar with MinGW, but I don't mind adding the sorting to both. Or directly in the `build_ext` command if the distutils maintainers think it would belong there. Just waiting on some answer on the PR itself by maintainers.

Avasam:
@abravalheri @jaraco Do you have a preference on whether this should be duplicated in the mingw compiler, or if it should go in build_ext ?

PR #372, opened 2025-05-31
Improve distutils.dist.Distribution from typeshed's setuptools.setup

Avasam:
Taken from https://github.com/python/typeshed/blob/9b7b2bafab7599195f400e1bf9f31620321aeb2e/stubs/setuptools/setuptools/__init__.pyi#L61-L105

These are used for https://github.com/pypa/setuptools/pull/5021

### DIFF
--- a/distutils/dist.py
+++ b/distutils/dist.py
@@ -13,7 +13,7 @@ import pathlib
 import re
 import sys
 import warnings
-from collections.abc import Iterable, MutableMapping
+from collections.abc import Iterable, Mapping, MutableMapping
 from email import message_from_file
 from typing import (
     IO,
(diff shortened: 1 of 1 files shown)


distutils/dist.py
```diff
@@ -223,17 +223,17 @@ def __init__(self, attrs: MutableMapping[str, Any] | None = None) -> None:  # no
...
         self.packages: list[str] | None = None
         self.package_data: dict[str, list[str]] = {}
-        self.package_dir: dict[str, str] | None = None
+        self.package_dir: Mapping[str, str] | None = None
         self.py_modules: list[str] | None = None
-        self.libraries = None
-        self.headers = None
+        self.libraries: list[tuple[str, dict[str, Any]]] | None = None
+        self.headers: list[str] | None = None
         self.ext_modules: list[Extension] | None = None
```
Avasam, line 228: typeshed typed this as `Sequence`, but as far as I can see
https://github.com/pypa/distutils/blob/603b94ec2e7876294345e5bceac276496b369641/distutils/command/build_ext.py#L178
and
https://github.com/pypa/distutils/blob/603b94ec2e7876294345e5bceac276496b369641/distutils/command/build_ext.py#L370-L383
means it *has* to be a list.

PR #371, opened 2025-05-25
Address ruff issues

DimitriPapadopoulos:

PR #373, opened 2025-07-11
Use dataclass to add type annotations to `Extension` so that it can easily be extended in `setuptools`.

abravalheri:
In https://github.com/pypa/setuptools/pull/5022, there is an effort to add type annotations to `setuptools.extension.Extension`.

However, right now it basically requires repeating all the `__init__` arguments that are already specified in `distutils.extension.Extension` and all the type annotations.
If any argument changes (added/removed/changed type annotation), that needs to be repeated in `setuptools`.

There is a discussion in https://github.com/pypa/setuptools/pull/5022/files#r2124429494 about how would it be possible to avoid this duplication and need for manual synchronisation.

This PR is an attempt to use `dataclass` for that.

I have experimented with other approaches (like attempting to use a TypedDict), but in the end of the day, this approach is what seems to me the most promising.

If I am not mistaken, this should allow backward compatibility. Type checkers are likely to start catching unknown keyword arguments, but I think that is a good thing.

On a related not, I did modify the warning to mention using unknown keywords is deprecated, the reasoning is that removing this lenient behaviour simplifies a lot of stuff and we can use dataclasses by the book.

The docstrings for the properties were basically moved from the existing class docstring.


abravalheri:
There seems to be an unrelated error in `[mypy] distutils/compilers/C/base.py`, possibly caused by recent versions of dependencies being released: see https://github.com/pypa/distutils/pull/374/files.


I have rebased this PR on top of #374, so that we can see if the main log fails on not in the majority of the CI workers. Now the same unrelated errors as in https://github.com/pypa/distutils/pull/374#issuecomment-3062772198 can be found.

distutils/compilers/C/base.py
```diff
@@ -70,7 +70,7 @@ class Compiler:
     # dictionary (see below -- used by the 'new_compiler()' factory
     # function) -- authors of new compiler interface classes are
     # responsible for updating 'compiler_class'!
-    compiler_type: ClassVar[str] = None  # type: ignore[assignment]
+    compiler_type: ClassVar[str] = None
```
abravalheri: See #374

distutils/extension.py
```diff
@@ -20,138 +22,161 @@
...
+    if not TYPE_CHECKING:
+
+        def __init__(self, *args, **kwargs):
+            extra = {repr(k): kwargs.pop(k) for k in tuple(kwargs) if k not in _safe}
+            if extra:
+                msg = f"""
+                Please remove unknown `Extension` options: {','.join(extra)}
+                this kind of usage is deprecated and may cause errors in the future.
+                """
+                warnings.warn(msg)
```
abravalheri: If/when lenient behaviour is not needed the whole custom `__init__` can be removed (and the classes `_Extension/Extension` can be unified.

distutils/extension.py
```diff
@@ -20,138 +22,161 @@
...
+    """
+
+    optional: bool = False
+    """
+    specifies that a build failure in the extension should not abort the
+    build process, but simply not install the failing extension.
+    """
+
+
+# Legal keyword arguments for the Extension constructor
```
abravalheri: The phrase Legal keyword arguments for the Extension constructorwas copied from `core.py` (not sure if `"legal"is the best term to use).

distutils/extension.py
```diff
@@ -20,137 +22,133 @@
...
-        if len(kw) > 0:
-            options = [repr(option) for option in kw]
-            options = ', '.join(sorted(options))
-            msg = f"Unknown Extension options: {options}"
-            warnings.warn(msg)
-
-    def __repr__(self):
-        return f'<{self.__class__.__module__}.{self.__class__.__qualname__}({self.name!r}) at {id(self):#x}>'
+
+# Legal keyword arguments for the Extension constructor
```
abravalheri: The phrase "Legal keyword arguments for the Extension constructor" was copied from `core.py` (not sure if `"legal"` is the best term to use).

distutils/_dataclass.py
```diff
@@ -0,0 +1,50 @@
...
+
+@dataclass_transform()
+def lenient_dataclass(**dc_kwargs):
+    """
+    Problem this class intends to solve:
+    - We need to modify __init__ so to achieve backwards compatibility
+      and keep allowing arbitrary keywords to be ignored
+    - But we don't want to throw away the dataclass-generated __init__
+      specially because we don't want to have to redefine all the typing
+      for the function arguments
```
abravalheri:
```suggestion
      for the function arguments
      
      If/when lenient behaviour and backward compatibility is not needed the whole
      customization can be removed, a regular ``dataclass`` can be used instead.
```

abravalheri:
I think that we could use the new [`dataclass_transform`](https://typing.python.org/en/latest/spec/dataclasses.html#the-dataclass-transform-decorator) to simplify even further the implementation.

abravalheri:
CI errors seem to indicate formatting changes (possibly related to new versions of linters available).

Avasam:
> I think that we could use the new [`dataclass_transform`](https://typing.python.org/en/latest/spec/dataclasses.html#the-dataclass-transform-decorator) to simplify even further the implementation.

Neat.

Overall I like how it also brings the doc closer to its declarations (and deduplicates type definitions).

Most of the added functional code is compat code that would later be removed.

> CI errors seem to indicate formatting changes (possibly related to new versions of linters available).

There's a pile of different CI issues at the moment, I think we're gonna need them merged with careful review knowing the CI will stay red until all fixed:
- https://github.com/jaraco/skeleton/pull/187
- #371 (part of the formatting issue you just identified)
- #374 or #368
- #383
- #392 (the other part of the formatting issue you just identified)

distutils/tests/test_extension.py
```diff
@@ -141,3 +144,59 @@ class setuptools_Extension(Extension):
     assert "_full_name" not in ext2.__dict__  # not initialized by default
     ext2._full_name = "hello world"  # can still be set in build_ext
     assert ext2._full_name == "hello world"
+
+
+TYPE_INFERENCE = {
+    # Simple example
+    """
+    from distutils.extension import Extension
+
```
Avasam, line 153: why not just... have this code be directly in a file that mypy runs on ? You can use `assert_type` instead of `reveal_type`.
For example: https://github.com/python/typeshed/blob/main/stubs/setuptools/%40tests/test_cases/check_setup.py

You can even test for errors, as long as "no-unecessary type ignore" is enabled:
https://github.com/python/typeshed/blob/main/stubs/setuptools/%40tests/test_cases/check_protocols.py

abravalheri:
I have also added `test_inference_sanity_check` to ensure the overall effect of adopting this approach results on the expected type inference.

distutils/tests/test_extension.py
abravalheri: Mostly because I don't like the idea of manually writing the full signature of `Extension.__init__` by hand, so that it can be added to the other side of the assertion/comparison :P

This way I can be lazy and just add some meaningful fragments.

And it does not matter much if mypy fails the typecheck due to other files in the project not being fully typed (right now I need to avoid the existing `mypy.ini`, otherwise `reveal_type` does not show anything)...

Avasam: Ah I glanced over the test too quickly and didn't see it was a dict, not a giant multiline string. Fair enough

distutils/tests/test_extension.py
```diff
@@ -106,12 +111,95 @@ def test_extension_init(self):
...
+        "swig_opts: builtins.list[builtins.str]",
+        "py_limited_api: builtins.bool",
+        "_full_name: builtins.str",
+    ],
+}
+
+
+@pytest.mark.filterwarnings("ignore::EncodingWarning")  # mypy.api.run
+@pytest.mark.parametrize("example,expectations", TYPE_INFERENCE.items())
+def test_inference_sanity_check(
```
Avasam, line 183: Should be skipped on pypy and cygwin/mingw. Or more generally, "should be skipped if mypy is not installed", which is already how the static types checkers tests work naturally with pytest plugin

PR #376, opened 2025-09-20
ClangCL compiler support

boberfly:
Added clang-cl support. I also have this at https://github.com/mnraker-pm/setuptools/tree/clangcl but saw that `distutils` exists separately and I guess is merged into `setuptools` from here.

I made this independently of https://github.com/pypa/distutils/pull/7 as I didn't see it until now.

The purpose/motivation at least for myself is to eventually cross-compile various projects from linux to windows via the clang-cl toolchain. I tested this with `markupsafe` from windows and clang-cl installed via vs_setup.

Added some basic tests to verify that the right executables are found and that the `compiler_type` is `clangcl` but this will more or less follow the same codepath as the `msvc` compiler.

With the `DISTUTILS_USE_SDK` environment variable enabled, the `VCTOOLSINSTALLDIR` and `WINDOWSSDKDIR` can be set externally from setuptools/distutils and this will be passed onto `clang-cl` and `lld-link`'s `/vctoolsdir` and `/winsdkdir` flags, ideal for cross-compiling on non-windows platforms.

### DIFF
--- a/distutils/_msvccompiler.py
+++ b/distutils/_msvccompiler.py
@@ -5,6 +5,7 @@ from .compilers.C import msvc
 __all__ = ["MSVCCompiler"]
 
 MSVCCompiler = msvc.Compiler
+ClangCLCompiler = msvc.ClangCLCompiler
 
 
 def __getattr__(name):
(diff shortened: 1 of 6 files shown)


MLopez-Ibanez:
I believe msvc cl.exe is miscompiling my python package on Windows ARM64 and I cannot find a way to use clang-cl instead of msvc to test this. I am using python-cffi which uses distutils.

MugundanMCW:
Hi @jaraco 
Any updates on this PR? this is really needed one, MSVC is being selected by default even after overriding the compiler env variables.

PR #379, opened 2025-10-26
Use float value of ST_MTIME in copy_file function

rajputa-deshaw:
The `copy_file` function uses `st[ST_MTIME]` when copying timestamps which is an int 
whereas the `newer` function uses `os.path.getmtime` which returns a float to check if
a file is outdated. This causes the copied file to always be older than the source file.

This commit updates the `copy_file` function to use `st.st_mtime` which is the float value.

Ref: pypa/setuptools#5079

### DIFF
--- a/distutils/file_util.py
+++ b/distutils/file_util.py
@@ -161,9 +161,9 @@ def copy_file(  # noqa: C901
         # According to David Ascher <da@ski.org>, utime() should be done
         # before chmod() (at least under NT).
         if preserve_times:
-            os.utime(dst, (st[ST_ATIME], st[ST_MTIME]))
+            os.utime(dst, (st.st_atime, st.st_mtime))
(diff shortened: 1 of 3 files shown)


rajputa-deshaw:
@Avasam Could you please approve the test workflow?

Avasam:
> @Avasam Could you please approve the test workflow?

Whilst I have contributed to, I am not a maintainer of distutils (or setuptools). I cannot approve PRs or workflows.

rajputa-deshaw:
@jaraco Could you review this PR?

rajputa-deshaw:
Actually, we should use `os.utime(dst, ns=(st.st_atime_ns, st.st_mtime_ns))` for full precision.

jaraco (approved):

PR #382, opened 2025-11-02
type `Command.force`

Avasam:
For https://github.com/pypa/setuptools/pull/4861

### DIFF
--- a/distutils/cmd.py
+++ b/distutils/cmd.py
@@ -107,7 +107,7 @@ class Command:
         # timestamps, but methods defined *here* assume that
         # 'self.force' exists for all commands.  So define it here
         # just to be safe.
-        self.force = None
+        self.force: bool | None = None
 
         # The 'help' flag is just used for command-line parsing, so
         # none of that complicated bureaucracy is needed.
--- a/distutils/cmd.py
+++ b/distutils/cmd.py
@@ -102,7 +102,7 @@ class Command:
         # timestamps, but methods defined *here* assume that
         # 'self.force' exists for all commands.  So define it here
         # just to be safe.
-        self.force = None
+        self.force: bool | None = None
 
         # The 'help' flag is just used for command-line parsing, so
         # none of that complicated bureaucracy is needed.
--- a/distutils/command/build_ext.py
+++ b/distutils/command/build_ext.py
@@ -125,7 +125,7 @@ class build_ext(Command):
         self.rpath = None
         self.link_objects = None
         self.debug = None
-        self.force = None
+        self.force: bool = None  # Should always be set in finalize_options
         self.compiler = None
         self.swig = None
         self.swig_cpp = None
--- a/distutils/command/build_ext.py
+++ b/distutils/command/build_ext.py
@@ -111,9 +111,9 @@ class build_ext(Command):
 
     def initialize_options(self) -> None:
         self.extensions = None
-        self.build_lib = None
-        self.plat_name = None
-        self.build_temp = None
+        self.build_lib: str = None  # Should always be set in finalize_options
+        self.plat_name: str = None  # Should always be set in finalize_options
+        self.build_temp: str = None  # Should always be set in finalize_options
         self.inplace = False
         self.package = None
 
--- a/distutils/command/build_ext.py
+++ b/distutils/command/build_ext.py
@@ -109,7 +109,7 @@ class build_ext(Command):
         ('help-compiler', None, "list available compilers", show_compilers),
     ]
 
-    def initialize_options(self):
+    def initialize_options(self) -> None:
         self.extensions = None
         self.build_lib = None
         self.plat_name = None
@@ -125,7 +125,7 @@ class build_ext(Command):
         self.rpath = None
         self.link_objects = None
         self.debug = None
-        self.force = None
+        self.force: bool = None  # Should always be set in finalize_options
         self.compiler = None
         self.swig = None
         self.swig_cpp = None
--- a/distutils/compilers/C/base.py
+++ b/distutils/compilers/C/base.py
@@ -70,7 +70,7 @@ class Compiler:
     # dictionary (see below -- used by the 'new_compiler()' factory
     # function) -- authors of new compiler interface classes are
     # responsible for updating 'compiler_class'!
-    compiler_type: ClassVar[str] = None  # type: ignore[assignment]
+    compiler_type: ClassVar[str] = None
 
     # XXX things not handled by this compiler abstraction model:
     #   * client can't provide additional options for a compiler,


distutils/command/build_ext.py
```diff
@@ -125,7 +125,7 @@ def initialize_options(self):
         self.rpath = None
         self.link_objects = None
         self.debug = None
-        self.force = None
+        self.force: bool = None  # Should always be set in finalize_options
```
Avasam, line 128: Not actually type-safe and easy to break in the future, but this is only place in distutils we assume a type for after `finalize_options`.

Alternatively this can be kept as  `bool | None`, which still works everywhere expecting a falsy/truthy value. But in some places where an actual boolean is expected (like https://github.com/pypa/setuptools/blob/d198e86f57231e83de87975c5c82bc40c196da79/setuptools/command/build_ext.py#L225-L227 ) we'd have to coerce to a boolean: `bool(self.force)`

distutils/compilers/C/base.py
```diff
@@ -70,7 +70,7 @@ class Compiler:
     # dictionary (see below -- used by the 'new_compiler()' factory
     # function) -- authors of new compiler interface classes are
     # responsible for updating 'compiler_class'!
-    compiler_type: ClassVar[str] = None  # type: ignore[assignment]
+    compiler_type: ClassVar[str] = None
```
Avasam: Type ignore is accurate, but technically we ignore all `assignment` atm, will be added back with https://github.com/pypa/distutils/pull/368

distutils/cmd.py
```diff
@@ -102,7 +102,7 @@ def __init__(self, dist: Distribution) -> None:
         # timestamps, but methods defined *here* assume that
         # 'self.force' exists for all commands.  So define it here
         # just to be safe.
-        self.force = None
+        self.force: bool | None = None
```
jaraco, line 105: How about instead a class property `force: bool = False`, allowing the option to be made True in instances? Then, give it a simple docstring "ignore file system timestamps". Then in build_ext, just remove the assignment in initialize options.

jaraco: I acknowledge - I'm violating my own policy of "don't change the implementation to accommodate typing", but now that we're getting pretty stable with typing, I'd like to start addressing the gaps implied by types reflected from a sloppy implementation.

jaraco: Okay. A simple default isn't possible because of the architecture of `set_undefined_options` relying on `None`, so there's nothing more to be done here.

Avasam: Since sometimes `None` is used to detect "not set by user" and follow different code paths, just need to make sure that wasn't the case here.

I don't think that's the case here. I think that, as you said, I tried to not change the implementation where not needed.

Falsy (`None` or `False`) checks should stay relatively safe. Could break a downstream edge case that was depending on this being explicitly `None`, falling back to a value that could come from elsewhere.

Up to you.

distutils/command/build_ext.py
```diff
@@ -109,11 +109,11 @@ class build_ext(Command):
...
 
-    def initialize_options(self):
+    def initialize_options(self) -> None:
         self.extensions = None
-        self.build_lib = None
-        self.plat_name = None
-        self.build_temp = None
+        self.build_lib: str = None  # Should always be set in finalize_options
+        self.plat_name: str = None  # Should always be set in finalize_options
+        self.build_temp: str = None  # Should always be set in finalize_options
```
jaraco, line 116: @Avasam This change was necessary to get the build to pass. Is this also acceptable?

Avasam: Not quite:
1. By adding `-> None` to `initialize_options`, mypy now parses and type-checks this function. So ALL members you initialize and don't explicitly type can only ever be `None`. (I've very purposefully avoided dealing with that until more of the project's typing could be cleared up) (surely one of the mypy PRs still left would've detected that issue)
2. The annoying thing with this initialize/finalize setup in general, is that you get false-positives comparing against `None` in `finalize_options`. Including user-defined overloads of `finalize_options`.
For now, I've been fixing types to `X | None` when it could still be `None` after `finalize_options`. And fixing bad `None` defaults when that was appropriate. The third case (as per point 1) I've mostly left alone for now ^^"

If you HAVE to decide, I'd say keeping the type accurate to AFTER running `finalize_options` is best, and it'll just be a known distutils/setuptools typing caveat that there will be false-positives when comparing against `None` in `initialize_options`/`finalize_options`. (so what you have right now, but type ALL members before adding `-> None` to `initialize_options`)

But all is not lost, there's one more trick you can use we developed in typeshed: `MaybeNone` aka "The `Any` Trick"
- https://typing.python.org/en/latest/guides/writing_stubs.html#the-any-trick
- https://github.com/python/typeshed/blob/eda55c02d0868770a06b6dad9f74ab608ca7799a/stdlib/_typeshed/__init__.pyi#L55-L59

(it'll obviously reduce type safety, but false-negatives are generally preferred over false-positives, especially for a legacy system like this)

Avasam: In case of doubt, let's keep working through other PRs and I can tackle this more in depth

jaraco: I'm not sure I follow, but this does allow the tests to pass, so let's go with it for now and we can refine subsequently. Thanks.

Avasam: See the new failing test results in https://github.com/pypa/distutils/pull/368 for what I meant

```
distutils/command/build_ext.py:114: error: Incompatible types in assignment (expression has type "None", variable has type "str")  [assignment]
distutils/command/build_ext.py:115: error: Incompatible types in assignment (expression has type "None", variable has type "str")  [assignment]
distutils/command/build_ext.py:116: error: Incompatible types in assignment (expression has type "None", variable has type "str")  [assignment]
distutils/command/build_ext.py:128: error: Incompatible types in assignment (expression has type "None", variable has type "bool")  [assignment]
distutils/command/build_ext.py:176: error: Incompatible types in assignment (expression has type "str | None", variable has type "None")  [assignment]
distutils/command/build_ext.py:178: error: Incompatible types in assignment (expression has type "list[Extension] | None", variable has type "None")  [assignment]
distutils/command/build_ext.py:185: error: Incompatible types in assignment (expression has type "list[str]", variable has type "None")  [assignment]
distutils/command/build_ext.py:206: error: Incompatible types in assignment (expression has type "list[Never]", variable has type "None")  [assignment]
distutils/command/build_ext.py:208: error: Incompatible types in assignment (expression has type "list[Never]", variable has type "None")  [assignment]
distutils/command/build_ext.py:213: error: Incompatible types in assignment (expression has type "list[Never]", variable has type "None")  [assignment]
distutils/command/build_ext.py:280: error: Incompatible types in assignment (expression has type "list[Never]", variable has type "None")  [assignment]
```

PR #383, opened 2025-11-02
Fix regex tests on Python 3.14

### LINKED ISSUE #381: Python 3.14 tests failing with `AssertionError: assert '(?s:foo[^/]*)\\z' == '(?s:foo[^/]*)\\Z'`
Recently noticed that all CI runs using Python 3.14 are failing with `AssertionError: assert '(?s:foo[^/]*)\\z' == '(?s:foo[^/]*)\\Z'`.

This includes the jobs `test (3.14, ubuntu-latest)` and `collateral (diffcov)`.

Complete error:
```py
_________________________ TestFileList.test_glob_to_re _________________________

self = <distutils.tests.test_filelist.TestFileList object at 0x7f78b4447390>

    def test_glob_to_re(self):
        sep = os.sep
        if os.sep == '\\':
            sep = re.escape(os.sep)
    
        for glob, regex in (
            # simple cases
            ('foo*', r'(?s:foo[^%(sep)s]*)\Z'),
            ('foo?', r'(?s:foo[^%(sep)s])\Z'),
            ('foo??', r'(?s:foo[^%(sep)s][^%(sep)s])\Z'),
            # special cases
            (r'foo\\*', r'(?s:foo\\\\[^%(sep)s]*)\Z'),
            (r'foo\\\*', r'(?s:foo\\\\\\[^%(sep)s]*)\Z'),
            ('foo????', r'(?s:foo[^%(sep)s][^%(sep)s][^%(sep)s][^%(sep)s])\Z'),
            (r'foo\\??', r'(?s:foo\\\\[^%(sep)s][^%(sep)s])\Z'),
        ):
            regex = regex % {'sep': sep}
>           assert glob_to_re(glob) == regex
E           AssertionError: assert '(?s:foo[^/]*)\\z' == '(?s:foo[^/]*)\\Z'
E             
E             - (?s:foo[^/]*)\Z
E             ?               ^
E             + (?s:foo[^/]*)\z
E             ?               ^

distutils/tests/test_filelist.py:64: AssertionError
```

Looks like https://docs.python.org/3/library/fnmatch.html#fnmatch.translate changed in an undocumented way in 3.14:
```
> uvx --python=3.13 python -c "import fnmatch; print(fnmatch.translate('foo*'))"
(?s:foo.*)\Z
> uvx --python=3.14 python -c "import fnmatch; print(fnmatch.translate('foo*'))"
(?s:foo.*)\z
```

Avasam:
Fixes #381 

This updates tests expectations, but alternatively I can update `glob_to_re`'s implementation to always capitalize `\z` so that the result stays consistent across Python versions (at least for as long as `\Z` stays supported, which I guess is gonna be quite a while)

### DIFF
--- a/distutils/compilers/C/base.py
+++ b/distutils/compilers/C/base.py
@@ -70,7 +70,7 @@ class Compiler:
     # dictionary (see below -- used by the 'new_compiler()' factory
     # function) -- authors of new compiler interface classes are
     # responsible for updating 'compiler_class'!
-    compiler_type: ClassVar[str] = None  # type: ignore[assignment]
+    compiler_type: ClassVar[str] = None
 
     # XXX things not handled by this compiler abstraction model:
     #   * client can't provide additional options for a compiler,
--- a/distutils/tests/test_filelist.py
+++ b/distutils/tests/test_filelist.py
@@ -3,6 +3,7 @@
 import logging
 import os
 import re
+import sys
 from distutils import debug, filelist
 from distutils.errors import DistutilsTemplateError
 from distutils.filelist import FileList, glob_to_re, translate_pattern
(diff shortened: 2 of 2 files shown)


distutils/tests/test_filelist.py
```diff
@@ -45,22 +46,22 @@ def assertWarnings(self, caplog):
         caplog.clear()
 
     def test_glob_to_re(self):
-        sep = os.sep
-        if os.sep == '\\':
-            sep = re.escape(os.sep)
+        sep = re.escape(os.sep) if os.sep == '\\' else os.sep
+        # https://docs.python.org/3/whatsnew/3.14.html#re
```
Avasam, line 50: Also ref https://github.com/python/cpython/issues/140922

distutils/tests/test_filelist.py
```diff
@@ -45,22 +46,22 @@ def assertWarnings(self, caplog):
         caplog.clear()
 
     def test_glob_to_re(self):
-        sep = os.sep
-        if os.sep == '\\':
-            sep = re.escape(os.sep)
+        sep = re.escape(os.sep) if os.sep == '\\' else os.sep
```
Avasam: Alternatively:
```suggestion
        sep = r"\\" if os.sep == "\\" else os.sep
```

or even:
```suggestion
        sep = re.escape(os.sep)
```

distutils/compilers/C/base.py
```diff
@@ -70,7 +70,7 @@ class Compiler:
     # dictionary (see below -- used by the 'new_compiler()' factory
     # function) -- authors of new compiler interface classes are
     # responsible for updating 'compiler_class'!
-    compiler_type: ClassVar[str] = None  # type: ignore[assignment]
+    compiler_type: ClassVar[str] = None
```
Avasam, line 73: Will be added back in https://github.com/pypa/distutils/pull/368 , but tests currently fails with this due to `unused type-ignore`

PR #389, opened 2025-12-19
Handle free-threaded builds in build_ext.get_libraries

### LINKED ISSUE #5126: [BUG] Fail to compile extensions on Windows free-threading non-MSVC
### setuptools version

setuptools == 80.9.0

### Python version

3.13t, 3.14t

### OS

Windows

### Description

[blis](https://github.com/explosion/cython-blis) uses setuptools to compile Cython extensions with cmake. Due to nuances around the pure-C libraries, it does not use MSVC, but instead it uses `new_compiler(plat="nt", compiler="unix")`.

Compilation fails on 3.13t and 3.14t:
```bash
 clang.exe [...] -lpython314 [...]
 LINK : fatal error LNK1181: cannot open input file 'python314.lib'
```

In some environments which for some reason deploy both python314.lib and python314t.lib, like `actions/setup-python`, the linking instead succeeds but then the interpreter segfaults on import.

The issue is caused by this code branch, which fails to account for "t" variants of python*.lib:
https://github.com/pypa/setuptools/blob/d198e86f57231e83de87975c5c82bc40c196da79/setuptools/_distutils/command/build_ext.py#L772-L782

### Expected behavior

On 3.14t clang must be passed `-lpython314t` instead of `-lpython314`

### Workaround
Subclass `build_ext` and override the method `build_ext.get_libraries` to make it return `["python314t"]` instead of `["python314"]`.

### How to Reproduce

```bash
pip install git+https://github.com/explosion/cython-blis.git@9dbe572cef57b394f97b3114a3e28ec04eb3a39e
```

mwilliamson:
This updates build_ext.get_libraries to handle free-threaded builds of Python when using non-MSVC compilers.

Previously, the non-free-threaded library would be used for free-threaded builds of Python. For instance, when attempting to build an extension for python314t, the extension would be linked against python314.

This fixes pypa/setuptools#5126.

I've manually tested this by using a similarly patched setuptools to successfully build jq.py wheels using mingw32 on Windows for python314t. While there are some automated tests for build_ext, my impression from a quick look was that those tests rely on being run in the relevant environment, rather than (for instance) mocking the environment, and therefore updating the tests to check this case isn't straightforward.

### DIFF
--- a/distutils/command/build_ext.py
+++ b/distutils/command/build_ext.py
@@ -770,13 +770,14 @@ class build_ext(Command):
             from .._msvccompiler import MSVCCompiler
 
             if not isinstance(self.compiler, MSVCCompiler):
-                template = "python%d%d"
-                if self.debug:
-                    template = template + '_d'
-                pythonlib = template % (
+                pythonlib = "python%d%d" % (
                     sys.hexversion >> 24,
                     (sys.hexversion >> 16) & 0xFF,
                 )
+                if is_freethreaded():
+                    pythonlib += 't'
+                if self.debug:
+                    pythonlib += '_d'
                 # don't extend ext.libraries, it may be shared with other
                 # extensions, it is a reference to the original list
                 return ext.libraries + [pythonlib]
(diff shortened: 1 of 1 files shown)


jaraco:
I don't understand the diffcov error:

distutils/command/build_ext.py (50.0%): Missing lines 787

Doesn't matter for this small change.

Issue #384, opened 2025-11-02
msys2 / mingw tests failing with SSL certificate issue


Avasam:
This is the full error from CI logs:
```py
Obtaining file:///D:/a/distutils/distutils
  Installing build dependencies: started
  Installing build dependencies: finished with status 'done'
  Checking if build backend supports build_editable: started
  Checking if build backend supports build_editable: finished with status 'done'
  Getting requirements to build editable: started
  Getting requirements to build editable: finished with status 'error'
  error: subprocess-exited-with-error
  
  Getting requirements to build editable did not run successfully.
  exit code: 1
  
  [88 lines of output]
  D:/a/_temp/msys64/tmp/pip-build-env-ezc5lgag/overlay/lib/python3.12/site-packages/setuptools_scm/git.py:202: UserWarning: "D:/a/distutils/distutils" is shallow and may cause errors
    warnings.warn(f'"{wd.path}" is shallow and may cause errors')
  Traceback (most recent call last):
    File "D:/a/_temp/msys64/ucrt64/lib/python3.12/urllib/request.py", line 1344, in do_open
      h.request(req.get_method(), req.selector, req.data, headers,
    File "D:/a/_temp/msys64/ucrt64/lib/python3.12/http/client.py", line 1338, in request
      self._send_request(method, url, body, headers, encode_chunked)
    File "D:/a/_temp/msys64/ucrt64/lib/python3.12/http/client.py", line 1384, in _send_request
      self.endheaders(body, encode_chunked=encode_chunked)
    File "D:/a/_temp/msys64/ucrt64/lib/python3.12/http/client.py", line 1333, in endheaders
      self._send_output(message_body, encode_chunked=encode_chunked)
    File "D:/a/_temp/msys64/ucrt64/lib/python3.12/http/client.py", line 1093, in _send_output
      self.send(msg)
    File "D:/a/_temp/msys64/ucrt64/lib/python3.12/http/client.py", line 1037, in send
      self.connect()
    File "D:/a/_temp/msys64/ucrt64/lib/python3.12/http/client.py", line 1479, in connect
      self.sock = self._context.wrap_socket(self.sock,
                  ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    File "D:/a/_temp/msys64/ucrt64/lib/python3.12/ssl.py", line 455, in wrap_socket
      return self.sslsocket_class._create(
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
[...]

PR #390, opened 2025-12-20
`spawn`/`call`: don't force param `cmd` to be mutable + allow more kwargs on msvc

Avasam:
This has 2 advantages:
- `cmd` doesn't have to be mutable. So `tuples` are usable
- It actually improves the generic type because `MutableSequence` is invariant. which can cause issues with union types since you need to match the exact full union type (some checkers do allow partial union match for inline/literal sequences)
  - Needed to resolve all mypy `[arg-type]`

I also improved the possible param types by using an overload on `search_path` (since `search_path` allows `bytes | str | PathLike[str]`, but not `PathLike[bytes]`)

Open question: Since `verbose` isn't used in `spawn`, maybe it's better to not even expose it as a possible kwargs in `Compiler` classes ? (no runtime change, just hiding it from the type annotation in wrapper methods)

---

Edit: Updated with https://github.com/pypa/distutils/commit/3f09fab3a3cfd1b2bb86066ad826a665e7712ebf

### DIFF
--- a/distutils/compilers/C/base.py
+++ b/distutils/compilers/C/base.py
@@ -10,7 +10,7 @@ import pathlib
 import re
 import sys
 import warnings
-from collections.abc import Callable, Iterable, MutableSequence, Sequence
+from collections.abc import Callable, Iterable, Sequence
 from typing import (
     TYPE_CHECKING,
     ClassVar,
@@ -39,6 +39,8 @@ from .errors import (
 )
 
 if TYPE_CHECKING:
+    from subprocess import _ENV
+
     from typing_extensions import TypeAlias, TypeVarTuple, Unpack
 
     _Ts = TypeVarTuple("_Ts")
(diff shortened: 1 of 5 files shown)


Avasam:
- `AssertionError` in test to be fixed by https://github.com/pypa/distutils/pull/383
- mypy failure on PyPy would be solved by https://github.com/jaraco/skeleton/pull/187
- `msys2_mingw` failure is https://github.com/pypa/distutils/issues/384
- `ci_setuptools` failure is https://github.com/pypa/distutils/issues/391 / solved by https://github.com/pypa/distutils/pull/392

distutils/compilers/C/base.py
```diff
@@ -70,7 +72,7 @@ class Compiler:
     # dictionary (see below -- used by the 'new_compiler()' factory
     # function) -- authors of new compiler interface classes are
     # responsible for updating 'compiler_class'!
-    compiler_type: ClassVar[str] = None  # type: ignore[assignment]
+    compiler_type: ClassVar[str] = None
```
Avasam: Will be added back in https://github.com/pypa/distutils/pull/368 , but tests currently fails with this due to `unused type-ignore`

Avasam:
Simplified a lot with https://github.com/pypa/distutils/commit/3f09fab3a3cfd1b2bb86066ad826a665e7712ebf

PR #392, opened 2025-12-20
ci_setuptools: Skip tests that distutils has no effect on

### LINKED ISSUE #391: `ci_setuptools` test should not run ruff on setuptools' repo
I can understand a type-checking check to see how changing distutils types affect setuptools. 

But I don't think it makes sense to check formatting (or single-file linting, which ruff currently is) for an external repo. Which can only lead to failures when updating the tool. Such as

https://github.com/pypa/distutils/actions/runs/20390360987/job/58598680308?pr=390
```
.tox/py/lib/python3.10/site-packages/pytest_ruff/__init__.py:133: RuffError
----------------------------- Captured stdout call -----------------------------
unformatted: File would be reformatted
   --> setuptools/tests/test_sdist.py:1:1
255 |             assert path not in manifest
256 | 
257 |     _INVALID_PATHS = {
    -         "must be relative": lambda: (
    -             os.path.abspath(os.path.join("sdist_test", "f.h"))
258 +         "must be relative": lambda: os.path.abspath(os.path.join("sdist_test", "f.h")),
259 +         "can't have `..` segments": lambda: os.path.join(
260 +             "sdist_test", "..", "sdist_test", "f.h"
261 |         ),
    -         "can't have `..` segments": lambda: (
    -             os.path.join("sdist_test", "..", "sdist_test", "f.h")
262 +         "doesn't exist": lambda: os.path.join(
263 +             "sdist_test", "this_file_does_not_exist.h"
264 |         ),
    -         "doesn't exist": lambda: (
    -             os.path.join("sdist_test", "this_file_does_not_exist.h")
    -         ),
    -         "must be inside the project root": lambda: (
    -             symlink_or_skip_test(
    -                 touch(os.path.join("..", "outside_of_project_root.h")),
    -                 "symlink.h",
    -             )
265 +         "must be inside the project root": lambda: symlink_or_skip_test(
266 +             touch(os.path.join("..", "outside_of_project_root.h")),
267 +             "symlink.h",
268 |         ),
269 |     }
270 | 
```

Avasam:
Closes https://github.com/pypa/distutils/issues/391

mypy could be re-enabled after https://github.com/pypa/setuptools/issues/4689

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -234,7 +234,9 @@ jobs:
       - name: Run setuptools tests
         run: |
           cd integration/setuptools
-          tox
+          # Skip tests that distutils has no effect on.
+          # Speedup CI and avoid unexpected failures due to updates.
+          tox -- -p no:ruff -p no:mypy
         env:
           VIRTUALENV_NO_SETUPTOOLS: null
 
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -237,7 +237,9 @@ jobs:
       - name: Run setuptools tests
         run: |
           cd integration/setuptools
-          tox
+          # Skip tests that distutils has no effect on.
+          # Speedup CI and avoid unexpected failures due to updates.
+          tox -- -p no:ruff -p no:mypy
         env:
           VIRTUALENV_NO_SETUPTOOLS: null
 


jaraco:
Ideally, it should be possible to run the full test suite - We want to prove that while Setuptools' CI is green, this job will also be green.  Of course, Setuptools' CI hasn't been green.

By disabling tests, we're limiting the benefit of that check. I'm not sure we want to start suppressing checks in order to accommodate (mask) downstream issues. There are many more checks that aren't directly impacted by distutils, so by making this change, we're setting a precedent that maybe we should skip other flaky or incidentally broken tests that don't exercise distutils.

It feels to me like instead we should just ensure that Setuptools is green, as we need to do that for its own sake. That said, I don't feel too badly about this change other than the tech debt it brings. I guess let's try it; we can always revert later if needed.

Avasam:
> We want to prove that while Setuptools' CI is green, this job will also be green.  Of course, Setuptools' CI hasn't been green.
> 
> By disabling tests, we're limiting the benefit of that check.

For these static tests, you still get that assurance.

- ruff: if the same set of rules pass on distutils code, it'll pass when synced to setuptools. Files are analysed indivudually and cannot affect each other. So Ruff version and configs are the only factor.
Especially given setuptools' ruff.toml specifically excludes `setuptools._distutils`.

- mypy: As a dependency, it *would* have made sense, but because of the distutils hack, type checkers don't even see `import distutils` as comming from `setuptools._distutils`. Unless:
  - You're using `types-setuptools` on Python 3.12+ or
  - https://github.com/pypa/setuptools/issues/4689 gets implemented

So you loose nothing, and gain from reduced randomly noisy CI. In other words, these tests truly had no value (negative value even).

PR #393, opened 2025-12-26
Prevent object file collisions in parallel extension builds

e-luks:
Parallel builds of extensions that share source files may write to the same object file paths under a common build directory, resulting in race conditions and non-deterministic build outputs.

Use a per-extension build directory to isolate object files and ensure deterministic, parallel-safe builds.

Bug: pypa/setuptools#3942

### Tests

I tested this change downstream at Gentoo after encountering https://bugs.gentoo.org/967476 and https://bugs.gentoo.org/945376 (packages https://pypi.org/project/librt and https://pypi.org/project/pymongo/). Since the build failures are encountered non-determinstically, i compiled both librt and pymongo as often as needed, capping the compilation attempts to 100.

- Before the change I encountered the build failures after at most 10 compilations
- After the change I didn't encounter the issue, compiling both librt and pymongo successfully 100 times.

### DIFF
--- a/distutils/command/build_ext.py
+++ b/distutils/command/build_ext.py
@@ -562,9 +562,12 @@ class build_ext(Command):
         for undef in ext.undef_macros:
             macros.append((undef,))
 
+        # Per-extension build dir to avoid conflicts in parallel builds.
+        ext_build_temp = os.path.join(self.build_temp, ext.name)
+
         objects = self.compiler.compile(
             sources,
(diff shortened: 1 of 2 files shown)


hector-cao:
Thanks @e-luks 

Hello I tested this fix on Ubuntu devel where we have python 3.14
This fix resolves the build issue we had for libvirt-python

Can anyone takes a look and see if we can get this fix in ?

PR #374, opened 2025-07-11
Remove unused type ignore mark

abravalheri:
In https://github.com/pypa/distutils/actions/runs/16223655098/job/45810230266?pr=373#step:6:140 (https://github.com/pypa/distutils/pull/373), there seems to be an unrelated type check error:

image

I believe this can be caused by an update in test dependencies (e.g. mypy 1.16.1 seems have been released in June).

PR #395, opened 2026-01-13
Don't automatically add PyInit_<name> to export symbols in Python 3.15+

### LINKED ISSUE #387: Avoid passing `/EXPORT:PyInit_*` to MSVC for Python 3.15+ (PEP 793)
Hello,

CPython 3.15 implements [PEP 793](https://peps.python.org/pep-0793/), which adds a new module entrypoint, `PyModExport_*`, as an alternative to the (now soft-deprecated) `PyInit_*`.

Distutils/Setuptools [adds `PyInit_*` to `export_symbols`](https://github.com/pypa/distutils/blob/603b94ec2e7876294345e5bceac276496b369641/distutils/command/build_ext.py#L729), and then, on Windows, [adds an `/EXPORT:PyInit_*` option to the linker](https://github.com/pypa/distutils/blob/603b94ec2e7876294345e5bceac276496b369641/distutils/compilers/C/msvc.py#L526).
This means that modules without a `PyInit_*` function will fail to build using Setuptools on Windows.

Luckily the workaround is easy: add a `PyInit_*` function (one that's backwards-compatible, or one that always raises, depending on desired behaviour 3.14 and below). However, it would be great if no workaround was necessary. Also, as of 3.15 alpha 2 there's no pressing reason for extensions to migrate to `PyModExport` (but this will probably change with stable ABI for free-threading).

The `/EXPORT` option should not be necessary. According to [MSVC docs](https://learn.microsoft.com/en-us/cpp/build/reference/export-exports-a-function?view=msvc-170), it's preferred to *use `__declspec(dllexport)` in the source code* instead. And CPython's [`PyMODINIT_FUNC`](https://docs.python.org/3/c-api/extension-modules.html#c.PyMODINIT_FUNC) macro does add the `__declspec(dllexport)`.

Would it be acceptable for setuptools to avoid adding `/EXPORT` (or auto-filling `export_opts`) when compiling for CPython 3.15+?

I'd like to solve any issues on the CPython side, ideally in the alpha period:
- If `PyMODINIT_FUNC` not adding `__declspec(dllexport)` in some cases, let's fix that
- `PyMODINIT_FUNC` being only *recommended* (IMO, “Porting to Python 3.15” in CPython's What's New can say that build tools are now not expected to handle this, and so `PyMODINIT_FUNC` or equivalent is needed in the source)

encukou:
Since PEP 793, it's valid for modules to not have a PyInit function. Telling MSVC to export a nonexistent function will make it fail.

The function should be exported in code, using `PyMODINIT_FUNC` (which adds `__declspec(dllexport)`, which is preferred over `/EXPORT` according to [Microsoft docs].)

[Microsoft docs]: https://learn.microsoft.com/en-us/cpp/build/reference/export-exports-a-function?view=msvc-170

Fixes: #387


distutils/command/build_ext.py
```diff
@@ -728,23 +728,31 @@ def get_ext_filename(self, ext_name: str) -> str:
...
-        the .pyd file (DLL) must export the module "PyInit_" function.
+        export.  This returns, and possibly updates, 'ext.export_symbols'.
+
+        On Python 3.14 and below (that is, before a new export hook name was added),
+        it adds "PyInit_" + module_name 'ext.export_symbols'.
+        Only relevant on Windows, where the .pyd file (DLL) must export the module
+        import hook function.
+
+        Since Python 3.15, don't add anything.
+        An export directive should be in the code itself.
```
rgommers: The original could possibly confuse the reader, since it could be read as
the package author having to do something - so try to rephrase:
```suggestion
        An export directive was added to the "PyMODINIT_FUNC" implementation
        for Python 3.15, which is all that is needed.
```

distutils/command/build_ext.py
```diff
@@ -728,23 +728,31 @@ def get_ext_filename(self, ext_name: str) -> str:
 
     def get_export_symbols(self, ext: Extension) -> list[str]:
         """Return the list of symbols that a shared extension has to
-        export.  This either uses 'ext.export_symbols' or, if it's not
-        provided, "PyInit_" + module_name.  Only relevant on Windows, where
-        the .pyd file (DLL) must export the module "PyInit_" function.
+        export.  This returns, and possibly updates, 'ext.export_symbols'.
+
+        On Python 3.14 and below (that is, before a new export hook name was added),
+        it adds "PyInit_" + module_name 'ext.export_symbols'.
```
rgommers:
```suggestion
        it adds "PyInit_" + module_name to 'ext.export_symbols'.
```

rgommers (commented):
This implementation looks good to me, for now I only have a few very small suggestions for the docstring. I still need to do some testing with a few real-world packages.

For other reviewers: this relies on https://github.com/python/cpython/pull/141672, which was merged only last week.

rgommers:
I tested this change with the `setuptools/c` test case in https://github.com/Quansight-Labs/stable-abi-testing (side note: not yet fully polished, but it's usable enough to derive test cases from now; it contains test packages in a _build-backends x binding-generators_ matrix). 

With the latest released version of `setuptools`, that test package fails to build with 3.15.0a3 - which is the most recent build available in GHA's `setup-python` - with ([CI log](https://github.com/rgommers/stable-abi-testing/actions/runs/20993785002/job/60345446318)):
```
LINK : error LNK2001: unresolved external symbol PyInit_limited
```

After applying the patch in this PR to `setuptools._distutils`, that error goes away and the package builds and passes its tests ([CI log](https://github.com/rgommers/stable-abi-testing/actions/runs/20994571639/job/60348060177)).

That repo uses test packages with `PyMODEXPORT_FUNC` rather than `PyMODINIT_FUNC`, so that does not yet prove that this also works with the latter.

I then started with `psutil`, tested the patch with its existing wheel build CI matrix for completeness - that passed of course, since 3.15 wheels are not included yet (no cibuildwheel support), and this PR is a no-op on older Python versions.

I then also added a new `psutil` CI job for 3.15, before figuring out that `psutil` depends on `pywin32` which doesn't have an sdist and doesn't build from source. So that's where I gave up.

Testing on 3.15-alpha is pretty cumbersome given the many packaging components and packages that don't yet support 3.15. That said, I'm confident that this PR works, and the testing I was able to do all worked as expected - so I'm +1 for merging this PR.

mgorny:
@encukou, could you do the reformatting, etc. to make CI pass?

Avasam:
The following should resolve most of the unrelated CI failures.
- https://github.com/jaraco/skeleton/pull/187
- #371
- #374 or #368
- #383
- #392

encukou:
There's also #371.

As far as I can see the failures are in code that this PR doesn't touch, due to updated mypy & ruff.
Should I pin the linters/checkers to a working version?

Avasam:
> Should I pin the linters/checkers to a working version?

From my **personal experience** with maintainers (namely @jaraco ), I don't think that's necessary, but I'm also obviously not talking for them. They prefer to not pin unless absolutely necessary to see new failures as soon as they arrive and to not deal with micro-managing pins on many dozen of projects that could conflict with a shared base (https://github.com/jaraco/skeleton/). In the same vein, whilst annoying, unrelated failures not caused by the current changes are often not blocking.

IMO, it could make sense for you to pin in your PR if it helps you ensure you haven't broken stuff. I'll likely end up unpinning in my own PRs and/or a maintainer may end up asking you to remove the pin once they finally get to review 🤷‍♀️

Up to you really, I'm fairly certain the only thing blocking review of your PR is maintainers' free time.

encukou:
> I'm fairly certain the only thing blocking review of your PR is maintainers' free time.

That's my impression as well.


If there's anything I can help with, let me know.

Avasam:
@encukou Previously identified CI failures have all been resolved (there's a new `ci_setuptools` failure, but it's not bringing down the entire CI)

---

@rgommers Concerning your previous comment, pywin32 has Python 3.15 wheels as of version b312

PR #400, opened 2026-02-20
install: Escape curly brackets in virtualenv path

helasraizam:
#### Summary of error
As mentioned in #281 if a virtual environment contains curly brackets (/path/{W}/project), setuptools fails with the error "ValueError: invalid variable 'W'", which is difficult to diagnose and is readily misattributed to the package being installed.

#### Under the hood
In `distutils.command.install`, `install.finalize_options()` calls `install.expand_basedirs()`, which expands `install.install_base`, `install.install_platbase`, and `install.root` paths using format_map to expand curly bracket variables supported in these paths through --user, --home, --prefix flags, sysconfig, etc..  Without these flags/configs,, sys.prefix and sys.exec_prefix are invoked, for which curly brackets in the path should be interpreted literally.

#### What this PR does
This PR escapes curly brackets when reading `sys.prefix` and `sys.exec_prefix` into `install.prefix` to prevent them from later being interpreted as variables, which resulted in the error.  It assumes that `install.prefix` is always formatted (through `util.subst_vars` and `format_map`) before usage, which appears to be the case.  It does not change the curly bracket expansion behavior of config and flag arguments.

#### Test/MWE
```
mkdir -p /tmp/{W}/project
cd /tmp/{W}/project
virtualenv venv
source venv/bin/activate
pip install -U pip wheel setuptools
pip install --no-build-isolation --no-binary :all: mysqlclient
```

The above fails with "ValueError: invalid variable 'W'" in the current main branch of setuptools and works if this patch is applied to setuptools.  Tests with tox on distutils and patched setuptools appeared to show no relevant errors.

This is my first PR, kindly treat with scrutiny and do not hesitate with feedback.

### DIFF
--- a/distutils/command/install.py
+++ b/distutils/command/install.py
@@ -19,7 +19,13 @@ from ..debug import DEBUG
 from ..errors import DistutilsOptionError, DistutilsPlatformError
 from ..file_util import write_file
 from ..sysconfig import get_config_vars
-from ..util import change_root, convert_path, get_platform, subst_vars
+from ..util import (
+    change_root,
+    convert_path,
+    escape_curly_brackets,
+    get_platform,
+    subst_vars,
+)
 from . import _framework_compat as fw
 
 HAS_USER_SITE = True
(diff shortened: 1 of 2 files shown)


MohammedAlkindi (commented):
The escape covers only the default-prefix path. On this head (`32186a14`), Windows, CPython 3.13.13, the other three branches of `finalize_other()` still raise:

`--user` gives `ValueError: invalid variable 'braceuser'`, `--home` gives `'bracehome'`, and an explicit `--prefix` gives `'braceprefix'`.

Line 600 escapes only when `self.prefix is None`, so a caller-supplied prefix skips it, and lines 593 and 597 assign `install_userbase` and `home` straight through. `subst_vars` runs afterwards from `expand_basedirs()`, so any of those reaching `install_base` fails the same way the issue describes.

The suite is unaffected by the change: the failing set at merge base `a1c0ebc1` and at this head are identical, one pre-existing MinGW test. Diffing against current `main` instead is misleading, the branch is 166 commits behind it.

PR #398, opened 2026-02-10
CI fixes

lazka:
Some small fixes here and there, see the commits for details.

Fixes #384

PR #402, opened 2026-03-17
Fix CI - MyPy and Ruff updates

DimitriPapadopoulos:
Fix issues with recent versions of MyPy and Ruff.

Also change current pre-commit message:
```
ruff (legacy alias)......................................................Passed
```
to:
```
ruff check...............................................................Passed
```

Contains an alternative to one of the fixes in #398 and #392. Not entirely sure about which is best.


Avasam:
I think if you rebase this PR, you'll find most of these to already have been fixed (there's been a round of CI fixes in the last week)

DimitriPapadopoulos:
Indeed, just a couple commits left.

Avasam:
Looks like this is now unrelated to Ruff or mypy, and more about removing py39 compat (now that Python 3.10 is the minimum)

DimitriPapadopoulos:
Indeed, but that's an actual ruff suggestion.

PR #409, opened 2026-07-05
Replace deprecated typing.AnyStr with a constrained TypeVar in util.change_root

jaraco:
`typing.AnyStr` is deprecated and slated for removal, so this replaces its use in `util.change_root` with an explicit `TypeVar("_StrBytes", str, bytes)` — preserving the same type-checking behavior (bound between `str` and `bytes`).

This originated as @sobolevn's [pypa/setuptools#5232](https://github.com/pypa/setuptools/pull/5232). Since `setuptools/_distutils` is vendored from here and overwritten on each sync, the `util.py` portion of that PR was routed upstream to this repo so it persists; the setuptools-owned parts (`glob.py`, `bdist_egg.py`) were merged there.

### DIFF
--- a/distutils/util.py
+++ b/distutils/util.py
@@ -17,7 +17,7 @@ import sys
 import sysconfig
 import tempfile
 from collections.abc import Callable, Iterable, Mapping
-from typing import TYPE_CHECKING, AnyStr
+from typing import TYPE_CHECKING, TypeVar
 
 from jaraco.functools import pass_none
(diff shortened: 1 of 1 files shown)


Avasam:
In https://github.com/pypa/distutils/pull/392 I disable static tests failing on setuptools code that distutils has no weight / effect upon. Which should resolve your failing tests here.

PR #412, opened 2026-07-10
mypy: fix all return-value+type-var

Avasam:


distutils/compilers/C/base.py
```diff
@@ -1114,9 +1114,10 @@ def library_filename(
         libname: str,
         lib_type: str = "static",
         strip_dir: bool = False,
-        output_dir: str | os.PathLike[str] = "",  # or 'shared'
+        output_dir: str | os.PathLike[str] | None = "",  # or 'shared'
     ):
-        assert output_dir is not None
+        if output_dir is None:
+            output_dir = ""
```
Avasam, line 1128: Feels odd allowing this to be `None` for type checking purposes, only to immediatly assert it can't be `None`.
If `assert output_dir is not None` is an invariant that must be kept, then `msvc.Compiler.create_static_lib` should check the return of `_fix_object_args` before passing it to `library_filename`

PR #415, opened 2026-07-19
Emit -O1 instead of bare -O in the Cygwin and MinGW compilers

### LINKED ISSUE #4873: mingw64 win32 compilation issue - does not accept "-O"
### setuptools version

76.0

### Python version

Python 3.12 and 3.13

### OS

Windows 11

This works fine for 64 bit compiling (win_amd64).  

### Description

When using the compiler=mingw64 option, the option for "-O" optimization is added automatically.  When using the 32 bit compiling mode (-m32), the compiler demands -Ox, with x being an integer or alternate characters, and halts abruptly with an errro.  This is not a problem with 64 bit compiling, and I was able to add an extra -O2 option by tweaking build_ext.

### How to Reproduce

Create a project that calls mingw64 with the option -m32 to generate 32 bit code.  Would be easier to just change -O to -O1, which is equivalent.

### Other detail

gcc -shared -O -Wall -IC:\progra~2\python\python313\include -IC:\progra~2\python\python313\Include -c pyrho.c -o build\temp.win32-cpython-313\Release\pyrho.o "-O2 -m32"
cc1.exe: error: argument to '-O' should be a non-negative integer, 'g', 's' or 'fast'
error: command 'C:\\Program Files\\mingw64\\bin\\gcc.EXE' failed with exit code 1

jaraco:
The Cygwin and MinGW compilers hardcoded a bare `-O` optimization flag. GCC treats `-O` and `-O1` as equivalent, but `cc1` rejects the bare form when targeting 32-bit code:

```
cc1.exe: error: argument to '-O' should be a non-negative integer, 'g', 's' or 'fast'
```

so any `-m32` build fails while 64-bit builds are unaffected. This emits `-O1` instead in both `MinGW32Compiler` and the shared Cygwin `Compiler` (which carries the identical flaw).

Fixes pypa/setuptools#4873. Part of the Phase 1 warm-up tracked in pypa/setuptools#5264 (filed as pypa/setuptools#5265).

This supersedes pypa/setuptools#5277, which fixed the mingw compiler in setuptools' vendored copy. That contribution is preserved here (via cherry-pick, retargeted to this upstream repo) and extended to cover the Cygwin compiler and to assert the positive `-O1` case in the regression test. Credit to @dchaudhari7177 for the original fix.

## Tests

`test_no_bare_optimization_flag` (gated only on cygwin, following `test_customize_compiler_with_msvc_python`) constructs a `MinGW32Compiler` on a stock build and asserts every compiler invocation carries `-O1` and no bare `-O` — so the regression is actually exercised in CI rather than skipped behind `is_mingw()`.

### DIFF
--- a/distutils/compilers/C/cygwin.py
+++ b/distutils/compilers/C/cygwin.py
@@ -256,9 +256,6 @@ class MinGW32Compiler(Compiler):
         if is_cygwincc(self.cc):
             raise Error('Cygwin gcc cannot be used with --compiler=mingw32')
 
-        # Use ``-O1`` rather than a bare ``-O``. The two are equivalent to GCC,
-        # but ``cc1`` rejects the bare form under ``-m32``.
-        # https://github.com/pypa/setuptools/issues/4873
         self.set_executables(
             compiler=f'{self.cc} -O1 -Wall',
             compiler_so=f'{self.cc} -shared -O1 -Wall',
--- a/distutils/compilers/C/cygwin.py
+++ b/distutils/compilers/C/cygwin.py
@@ -256,11 +256,14 @@ class MinGW32Compiler(Compiler):
         if is_cygwincc(self.cc):
             raise Error('Cygwin gcc cannot be used with --compiler=mingw32')
 
+        # Use ``-O1`` rather than a bare ``-O``. The two are equivalent to GCC,
+        # but ``cc1`` rejects the bare form under ``-m32``.
+        # https://github.com/pypa/setuptools/issues/4873
         self.set_executables(
-            compiler=f'{self.cc} -O -Wall',
-            compiler_so=f'{self.cc} -shared -O -Wall',
-            compiler_so_cxx=f'{self.cxx} -shared -O -Wall',
-            compiler_cxx=f'{self.cxx} -O -Wall',
+            compiler=f'{self.cc} -O1 -Wall',
+            compiler_so=f'{self.cc} -shared -O1 -Wall',
+            compiler_so_cxx=f'{self.cxx} -shared -O1 -Wall',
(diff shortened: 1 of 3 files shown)


distutils/compilers/C/tests/test_mingw.py
```diff
@@ -20,12 +20,28 @@ def test_set_executables(self, monkeypatch):
...
+        assert compiler.compiler_cxx == split_quoted('c++ -O1 -Wall')
         assert compiler.linker_exe == split_quoted('cc')
         assert compiler.linker_so == split_quoted('cc -shared')
 
+    @pytest.mark.skipif('sys.platform == "cygwin"')
+    def test_no_bare_optimization_flag(self):
+        # A bare ``-O`` is rejected by cc1 under ``-m32``; an explicit level
+        # (``-O1``) is equivalent and accepted.
+        # https://github.com/pypa/setuptools/issues/4873
+        compiler = cygwin.MinGW32Compiler()
```
jaraco: Let's make this a docstring, describing what's being asserted, followed by this context.

distutils/compilers/C/cygwin.py
```diff
@@ -256,11 +256,14 @@ def __init__(self, verbose=False, force=False):
         if is_cygwincc(self.cc):
             raise Error('Cygwin gcc cannot be used with --compiler=mingw32')
 
+        # Use ``-O1`` rather than a bare ``-O``. The two are equivalent to GCC,
+        # but ``cc1`` rejects the bare form under ``-m32``.
+        # https://github.com/pypa/setuptools/issues/4873
```
jaraco: This concern is largely redundant with the regression test, so let's omit it, but extend the regression test docstring to capture the details about `GCC` and `cc1`.

PR #416, opened 2026-07-19
Use module-level named loggers in the C compiler modules

jaraco:
The C compiler modules (`base`, `msvc`, `unix`) imported the distutils **root** logger via `from ..._log import log` (`distutils._log.log = logging.getLogger()`). This replaces that with a module-level `log = logging.getLogger(__name__)` in each.

Named loggers still propagate to the root logger's handlers and inherit its effective level (distutils sets the root level by verbosity in `dist.py`), so message filtering and output are unchanged — but the compilers package no longer reaches into `distutils._log`. That's one more internal dependency removed on the path to a standalone `compilers.C`.

Part of Phase 1 in pypa/setuptools#5264 (filed as pypa/setuptools#5266).

## Verification

- `pytest distutils/compilers/C/tests/` — 18 passed, 13 skipped.
- No test asserts on these compiler log records; `caplog`-style captures at the root logger continue to see them via propagation regardless.
- `ruff check` clean on the three modules.


distutils/compilers/C/base.py
```diff
@@ -37,6 +37,8 @@
     UnknownFileType,
 )
 
+log = logging.getLogger(__name__)
```
jaraco: This name is going to migrate over time. It'll be `distutils.compilers.C.*` here, but `setuptools._distutils.compilers.C` when it moves to Setuptools, and later `compilers.C.*` when the package becomes its own package. I'm wondering if for the sake of stability if we should use `compilers.C.*` now. Yes, let's do that.

PR #417, opened 2026-07-20
Decouple the C compilers from distutils build/filesystem helpers

jaraco:
Removes the C compilers' dependence on distutils' build/filesystem helpers, per the analysis in pypa/setuptools#5267 (Phase 1 of pypa/setuptools#5264).

## What changed

**Vendored into the `compilers` namespace package** (stdlib-only, siblings of the `logging.py` from #5266):
- `compilers/_modified.py` — `newer`/`newer_group` mtime helpers.
- `compilers/_util.py` — `split_quoted`, vendored verbatim (a switch to `shlex.split` is not behavior-equivalent and is left for later).

**Subprocess execution reworked** into a modern `call` + deprecated `spawn`:
- New `Compiler.call(cmd, *, env=None, **kwargs)` — the modern entry point: logs the command, injects the macOS deployment target, and runs `subprocess.check_call`, letting **native** `subprocess` exceptions propagate.
- `Compiler.spawn` is now a **deprecation shim** — it warns, delegates to `call`, and translates `subprocess`/`OSError` failures to `DistutilsExecError` (imported late, to keep the package off `distutils.errors` at module scope).
- Internal call sites across `unix`/`cygwin`/`zos`/`msvc` moved to `call()` and now catch `subprocess.CalledProcessError`/`OSError`. MSVC's PATH injection moved from a `spawn` override to a `call` override.
- The macOS env injection moved out of `distutils.spawn` into `compilers/platform/macos.py` (`_inject_ver`) — the compiler path is the only place it applies (verified: no non-compiler `spawn` call site invokes a compiler/linker).
- The previously-vendored `compilers/spawn.py` is removed.

**`distutils.spawn.spawn` reduced to a thin wrapper** — it now only logs, forwards `cmd`/`**kwargs` to `subprocess.check_call`, and maps `OSError`/`CalledProcessError` to `DistutilsExecError`; no `shutil.which` (subprocess searches `PATH` itself) and no macOS injection. `Command.spawn` no longer forwards the now-defunct `search_path`.

**Modernized in place** — the public `Compiler` methods keep their signatures for backward compatibility, but their bodies now use the stdlib:
- `mkpath` → `os.makedirs(name, mode, exist_ok=True)` (empty-name guarded).
- `move_file` → `shutil.move`.
- `execute` → trivial log-and-call, no longer used internally.

**Eliminated** — the cygwin `.def`-file write is inlined (`pathlib.Path.write_text`), removing `distutils.file_util.write_file` and the `execute` indirection. Also removed the MSVC `_fallback_spawn` monkeypatch shim (numpy.distutils <1.19, per pypa/distutils#15) and its now-dead pytest filter.

Net: `compilers/C/*` no longer import `distutils.spawn`, `dir_util`, `file_util`, `_modified`, or `util.execute`/`split_quoted`.

## Deliberately out of scope

Two transitive couplings remain, by design, as they belong to sibling sub-issues:
- the vendored `_modified` / deprecated `spawn` still reference `Distutils*Error` (native `compilers.errors` is #5270 / S6);
- `_inject_ver` still calls `distutils.util` for the target-version helpers (platform/util decoupling is #5268 / S4).

## Verification

- Full distutils suite: 246 passed, 22 skipped. Compiler + build_ext + spawn subset green.
- Verified `call()` raises native exceptions without warning; `spawn()` warns and translates to `DistutilsExecError`; `distutils.spawn.spawn` still translates.
- `ruff check --select F,I` clean on all changed files.


distutils/compilers/C/base.py
```diff
@@ -21,16 +23,15 @@
...
     DistutilsPlatformError,
 )
-from ...file_util import move_file
-from ...spawn import spawn
-from ...util import execute, is_mingw, split_quoted
+from ...util import is_mingw
+from .._modified import newer_group
+from .._util import split_quoted
 from ..logging import get_logger
+from ..platform.macos import _inject_ver
```
jaraco: Use `from ..platform import macos`, so that the `macos` name is reflected in the call below.

distutils/compilers/C/base.py
```diff
@@ -1153,12 +1155,48 @@ def execute(
...
+        try:
+            self.call(cmd, env=env, **kwargs)
+        except OSError as exc:
+            raise DistutilsExecError(
+                f"command {cmd[0]!r} failed: {exc.args[-1]}"
+            ) from exc
+        except subprocess.CalledProcessError as err:
+            raise DistutilsExecError(
+                f"command {cmd[0]!r} failed with exit code {err.returncode}"
+            ) from err
```
jaraco: Let's unify this behavior with distutils.spawn.spawn, either by extracting a context manager to do the exception translation or by having "inner" spawn that's called after the logging is done.

PR #418, opened 2026-07-21
add compresslevel to archive_util.make_tarball, fix test_archive_util

ds-cbo:
title mostly speaks for itself, but as for the why: I notice that `build -s` is taking a very long time and 90% of it is spent hugely optimizing the tarball as level 9 is the default, I would like to fine-tune that in setuptools, but first distutils needs to support passing it along to `tarfile.open`

while testing, I noticed that the entire `tets_archive_util` file was ignored when running `python3.14 -m pytest`, which seemed to be fixed when prefixing the class with Test like any other class in the directory, presumably some leftover from a different testing framework

### DIFF
--- a/distutils/tests/test_archive_util.py
+++ b/distutils/tests/test_archive_util.py
@@ -4,6 +4,7 @@ import functools
 import operator
 import os
 import pathlib
+import shutil  # noqa: F401  # import used by skipIf
 import sys
 import tarfile
 from distutils import archive_util
@@ -17,6 +18,8 @@ from distutils.archive_util import (
 from distutils.spawn import spawn
 from distutils.tests import support
 from os.path import splitdrive
+from time import perf_counter
+from unittest import TestCase
 
 import path
 import pytest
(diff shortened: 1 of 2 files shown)


distutils/tests/test_archive_util.py
```diff
@@ -46,7 +49,7 @@ def same_drive(*paths):
     return all_equal(pathlib.Path(path).drive for path in paths)
 
 
-class ArchiveUtilTestCase(support.TempdirManager):
+class TestArchiveUtil(TestCase, support.TempdirManager):
```
ds-cbo, line 52: class name must start with Test for pytest to detect it as something that is testable, `TestCase` must be added for the default `addCleanup` function that `patch` is using

distutils/tests/test_archive_util.py
```diff
@@ -87,7 +102,7 @@ def test_make_tarball_extended(self):
 
     def _make_tarball(self, tmpdir, target_name, suffix, **kwargs):
         tmpdir2 = self.mkdtemp()
-        if same_drive(tmpdir, tmpdir2):
+        if not same_drive(tmpdir, tmpdir2):
```
ds-cbo, line 105: this looked like a logic error to me, it would skip every test

PR #421, opened 2026-07-31
Decouple the compilers from distutils version/platform helpers

jaraco:
Removes the compilers' dependence on distutils' version, platform, macOS, and linker-arg helpers, per pypa/setuptools#5268 (Phase 1 of pypa/setuptools#5264). Four commits, one per concern.

## What changed

- **Version** — `CygwinCCompiler.gcc_version` returns a `packaging.version.Version("11.2.0")` instead of the deprecated `distutils.version.LooseVersion` (and its `suppress_known_deprecation` guard).
- **Linker args** — `consolidate_linker_args` vendored from `distutils.compat` into `compilers/_util.py`.
- **macOS** — `compiler_fixup` (wrapping stdlib `_osx_support`) and the deployment-target logic (`get_macosx_target_ver` and friends) vendored into `compilers/platform/macos.py`. The deployment-target mismatch now raises `compilers.errors.PlatformError`.
- **Platform detection** — `get_platform`/`get_host_platform`/`is_mingw` vendored into `compilers/platform/detect.py` (host delegates to stdlib `sysconfig.get_platform()`); `base.py`/`msvc.py` point at them.

Net: `compilers/C/*` no longer import `distutils.util`, `distutils.version`, `distutils.compat`, or `distutils._macos_compat`.

## Behavior change

The macOS deployment-target-too-low error is now compiler-raised (`compilers.errors.PlatformError`) rather than framework-raised (`DistutilsPlatformError`) — a natural extension of #5270. distutils' top-level `except (DistutilsError, CCompilerError)` handlers still catch it (via `CCompilerError` = `compilers.errors.Error`). Updated the two `test_build_ext` deployment-target tests accordingly.

## Deliberately out of scope

Remaining `distutils` touches in the compilers, left for their own steps:
- `sysconfig` (`cygwin.py`, `macos.py`, `unix.py`) → the sysconfig decoupling, #5269 (S5). macOS config-var lookups intentionally still route through distutils.sysconfig so the source stays consistent until S5 migrates them together.
- the deprecated `Compiler.spawn` shim (`distutils.spawn._translate_errors`) — legacy by design.
- `distutils.debug.DEBUG` (debug-print flag) and `distutils.fancy_getopt.FancyGetopt` (`--help-compiler` formatting) — not version/platform shims; a separate follow-up.

## Verification

- Full `tox` (ruff + mypy + tests): 570 passed, 20 skipped, `py: OK`.


distutils/compilers/_util.py
```diff
@@ -4,6 +4,23 @@
...
+def consolidate_linker_args(args: _IterableT) -> _IterableT | str:
+    """
+    Ensure the return value is a string for backward compatibility.
+
+    Retain until at least 2025-04-31. See pypa/distutils#246
+    """
+
+    if not all(arg.startswith('-Wl,') for arg in args):
+        return args
+    return '-Wl,' + ','.join(arg.removeprefix('-Wl,') for arg in args)
```
jaraco: This function was previously in a `compat` module, suggesting it should be unneeded or inlined at some point... and the "some point" appears to be in the past (2025-04-31).

Let's figure out what the long-term implementation should be (address the 'backward-compatibility' and 'retain until' concerns and replace with a docstring explaining the intent). If it's to be retained, let's make it private ('_consolidate_linker_args').

Also, let's remove the version found in distutils.compat as it's no longer used.

distutils/compilers/platform/macos.py
```diff
@@ -4,25 +4,86 @@
...
+def _target_ver_from_syscfg():
+    """The deployment target latched into the interpreter's configuration."""
+    global _syscfg_target_ver
+    if _syscfg_target_ver is None:
+        # sysconfig remains distutils' pending the sysconfig decoupling
+        # (pypa/setuptools#5269).
+        from ... import sysconfig
+
+        _syscfg_target_ver = sysconfig.get_config_var(VERSION_VAR) or ''
+    return _syscfg_target_ver
```
jaraco, line 44: Looks like a lot of this was copied from the distutils util module. Let's consider consolidating the duplication; if distutils still requires this functionality, can it call it from the compilers package?

distutils/compilers/platform/detect.py
```diff
@@ -0,0 +1,40 @@
+"""Platform identification used to select and configure compilers."""
```
jaraco, line 1: This module introduces some duplication; can it be consolidated (re-used in distutils)?

distutils/compat/__init__.py
```diff
@@ -1,18 +1 @@
 from __future__ import annotations
```
jaraco, line 1: Probably this file should just be deleted now that it has no meaningful content.

jaraco: I'd prefer to keep it: `distutils/compat/` still holds the `numpy` and `py310` submodules, and retaining `__init__.py` keeps `distutils.compat` a regular package. Removing it would turn it into an implicit namespace package, which isn't the intent here.

PR #430, opened 2026-09-10
Adopt the standalone compilers packages

jaraco:
Adopt the standalone compilers distributions that were extracted from distutils, completing the extraction side of the compilers-decoupling epic (pypa/setuptools#5264, pypa/setuptools#5274).

## What

- Declare runtime dependencies on `compilers.C`, `compilers.common`, and `compilers.errors` (published from coherent-oss).
- Delete the vendored `distutils/compilers/` tree (its contents now live in, and are CI-tested by, those packages).
- Rewire the ~10 internal references from `distutils.compilers.*` to the top-level packages:
  - the compiler shims (`ccompiler`, `unixccompiler`, `_msvccompiler`, `cygwinccompiler`, `zosccompiler`) → `compilers.C.*`
  - `errors` → `compilers.C.errors` + `compilers.errors`
  - `util` → `compilers.common.platform`
  - `compat.numpy` → `compilers.C.base`
- The compiler tests move out with the tree; `conftest.py` drops the now-orphaned `suppress_path_mangle` fixture and repoints `disable_macos_customization` at `compilers.common.platform.macos`.

The public shims are unchanged in behavior — `distutils.ccompiler.CCompiler` is `compilers.C.base.Compiler`, `distutils.errors.CCompilerError` is `compilers.errors.Error`, `distutils.util.get_platform()` routes through `compilers.common.platform.detect`, etc.

There's no dependency cycle: `compilers.C` imports `distutils` only lazily on deprecated paths (deprecated `Compiler.spawn`, `debug_print`) and deliberately doesn't declare it (coherent-oss/compilers.C#1), so this is a clean one-way `distutils → compilers.C` dependency.

## Verification

With the three packages installed, the full suite passes (220 passed, 9 skipped) and ruff is clean.

## Notes

- No newsfragment here by design — the changelog entry will be added when this syncs into setuptools.
- **Companion setuptools work is required**: setuptools vendors distutils as `setuptools/_distutils`, so once distutils depends on `compilers.C`, setuptools must bundle or depend on `compilers.C`/`compilers.common`/`compilers.errors` for its `_distutils` to keep working. That's the pivot into Phase 4 (pypa/setuptools#5273) and will be handled separately.


jaraco:
CI is green except the expected `ci_setuptools` integration failure:

```
setuptools/_distutils/errors.py: from compilers.C.errors import ...
ModuleNotFoundError: No module named 'compilers'
```

That job drops the current distutils into setuptools' vendored `_distutils`, whose (synced) `errors.py` now imports from `compilers.C` — but setuptools' environment doesn't have the `compilers.*` packages installed. Resolving it is the companion setuptools change (bundle or depend on `compilers.C`/`compilers.common`/`compilers.errors`), i.e. Phase 4 of pypa/setuptools#5264 (pypa/setuptools#5273). Everything else — full test matrix, mypy, docs, coverage — passes.

